PR #5, opened 2019-05-07
Add a method to download licenses

carmenbianca:
A naive implementation that works fairly well.

I still need to extract some functionality from reuse._main.py::download into a separate function, so that a command such as `reuse init` doesn't need to reimplement anything.


carmenbianca:
Oh, I also forgot to write a "congratulations, successfully downloaded license" message.

PR #11, opened 2019-05-16
Add a method to add headers to files

carmenbianca:
This is really, really primitive right now, but it kind of works.

I opted for `reuse addheader` instead of `reuse header add`. I can see the appeal of the latter, but it's a little clunkier to implement, and I am not entirely certain what commands `reuse header` would have other than `add`. Even so, it would be non-non-trivial to change.

I am running into a couple of design issues, though. I reckon I'll detail them in issues later. They're hard for me to describe succinctly.

I am completely ignoring templates in this PR. They're for later.


carmenbianca:
Wrote design challenges in #12 and #13.

PR #14, opened 2019-05-21
Implement PathType to replace argparse.FileType

### LINKED ISSUE #10: Refactor out argparse.FileType
Some argparse arguments currently use `argparse.FileType`. This is really convenient because it automatically detects whether a file exists and/or is writeable. But it's also annoying because it creates a stream and doesn't handle directories.

It shouldn't be too hard to homebrew something for this that just returns a Path object.

carmenbianca:
Fixes #10

### DIFF
--- a/src/reuse/_util.py
+++ b/src/reuse/_util.py
@@ -11,6 +11,7 @@ import os
 import re
 import shutil
 import subprocess
+from argparse import ArgumentTypeError
 from gettext import gettext as _
 from hashlib import sha1
 from os import PathLike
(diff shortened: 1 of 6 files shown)


carmenbianca:
Bah, this is currently breaking exclusively in CI infrastructure. Maybe I can allow those tests to fail, or figure out another way to write those tests.

carmenbianca:
The tests fail because root can read files that it has no permissions to. Because of course root can do that. The CI is run as root (or at least, the Python Docker image is), and I don't know how to change that, so I'll just xfail those tests.

PR #19, opened 2019-05-30
Update directory name in README.rst

maxwell-k:
`git clone git@github.com:fsfe/reuse-tool.git` creates a directory called `reuse-tool/` not `reuse/`


carmenbianca:
Thanks @maxwell-k for spotting that!

PR #20, opened 2019-06-02
Amend docstring for files_without_copyright

maxwell-k:
Before this change the docstring for files_without_copyright  talks about license information. After this change it refers to copyright as expected.


carmenbianca:
You're absolutely right. Merging. Thank you!

PR #24, opened 2019-06-11
Forward a posix-like path to debian.copyright

### LINKED ISSUE #22: Dep5 should use OS-independent dir separators
Given `.reuse/dep5` file
```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/

Files: gradle/*.lockfile
Copyright: ©  Basil Peace
License: Apache-2.0
```
`reuse lint` (`master` branch) correctly recognizes license on `gradle/main.lockfile` file under Debian, but not under Windows.
Under Windows this works:
```
Files: gradle\\*.lockfile
```
which is somewhat strange.
Format should be OS-independent otherwise cross-platform development and build is not feasible.

carmenbianca:
str(Path) uses `\\` separators instead of `/`. Path.as_posix() returns
the path with separators that are expected by debian.copyright.

Reported-by: Basil Peace

Fixes #22

### DIFF
--- a/src/reuse/_util.py
+++ b/src/reuse/_util.py
@@ -184,7 +184,7 @@ def _determine_license_path(path: PathLike) -> Path:
 
 def _copyright_from_dep5(path: PathLike, copyright: Copyright) -> SpdxInfo:
     """Find the reuse information of *path* in the dep5 Copyright object."""
-    result = copyright.find_files_paragraph(str(path))
+    result = copyright.find_files_paragraph(Path(path).as_posix())
 
     if result is None:
         return SpdxInfo(set(), set())


carmenbianca:
Wasn't an upstream bug at all. Just unexpected behaviour from pathlib. It was, however, a trivial fix :) Or at least, I think it is. I have no Windows machine with which to test this.

@grv87 Would it be possible for you to test this branch? (windows-path)

grv87:
@carmenbianca, it works.
Thanks for quick fix!

carmenbianca:
Thanks again @grv87! Merging

PR #37, opened 2019-07-05
Dockerize tool

mxmehl:
This PR add a Dockerfile to install reuse in a Docker container based on Alpine. This way, it is easy to integrate in CI processes.

I deliberately did not choose to just install via pip to reduce the dependency on this source, and to be able to build the image and publish it to Docker Hub without the need to have a release on pypi.

Downside to far: The image is quite heavy with 325MB. The install via pip in the same base image results in an image size of ~60MB. I tried `make clean` to get rid of build artifacts, but afterwards the binary didn't run any more. Ideas?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+# SPDX-Copyright: 2019 Free Software Foundation Europe e.V.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+FROM alpine:latest
+
+# Dependencies for reuse-tool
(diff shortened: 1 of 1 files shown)


carmenbianca:
> The image is quite heavy with 325MB. The install via pip in the same base image results in an image size of ~60MB.

Can you show the Dockerfile of the pip version?

mxmehl:
Fixed thanks to Carmen's help. Local size is at 76MB now

PR #48, opened 2019-07-10
Show version of REUSE Spec in tool

### LINKED ISSUE #38: Specify what version of the spec the tool applies to

carmenbianca:
Fixes #38


mxmehl:
LGTM. What do you think about adding a string like "Your project is not compliant with version XX of the REUSE Specification" if lint fails? This way, we would help people understand if their lint suddenly fails if we introduce breaking changes.

carmenbianca:
Sure, that makes a lot of sense to me!

carmenbianca:
Added the message, merging.

PR #49, opened 2019-07-10
Implement `reuse download --all`

### LINKED ISSUE #45: Download all used licenses at once
`reuse download` could use an option to download all used licenses in a project, instead of having to download them individually. Perhaps with an `--all` flag?

Since this might cause some issues with badly formed license information and/or false-positive findings, this could also happen interactively.

carmenbianca:
Here's an extremely broken implementation that fixes #45.

The output is currently completely broken (i.e., prints a lot of error messages), but the functionality is there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ following sections:
 
 The versions follow [semantic versioning](https://semver.org).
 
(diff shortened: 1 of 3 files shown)


carmenbianca:
Blocked by #11, which introduces some functionality that this PR also requires

:-(

PR #64, opened 2019-08-09
Add new comment styles

kirelagin:
* TeX:
  * LaTeX
  * Erlang
* ML:
  * OCaml
  * SML

(Continuation of #63)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -20,6 +20,16 @@ Contributors
 
 - Kirill Elagin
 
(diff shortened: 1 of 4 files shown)


kirelagin:
(I have to admit that I did not test any of this because I’m not to good at running Python programs, and I was hoping that a CI will do my job here 🙄.)

carmenbianca:
Thank you very much! The CI now does the job of checking it, because I hadn't written a generic test yet :p But it all checks out. Thanks again :-)

Issue #68, opened 2019-08-13
Add configuration for addheader and provide --defaults for addheader


carmenbianca:
See also #13.

`git config` and `.reuse/config` should provide default values for author/email and default license respectively.

It makes sense for me to implement this as a composite of Project. This might be a little difficult, though, because I'd have to forward the config object to a lot of places that do not currently have one.

carmenbianca:
Additionally, make it possible to configure a default template for a given template.

nicorikken:
@carmenbianca @siiptuo @CharString @mxmehl 

I really like to see this feature become a reality, and at the moment there are already 2 good PR's to read copyright information from the project: #240 and #345 As these two are already in conflict, as they both add a single side-step to the explicit arguments, I think we need to discuss how this should be implemented. Also considering that in line with the `pyproject.toml` as a source in #345 there might be many more formats coming in the future. So let me share my thoughts, and hopefully we can refine through a discussion here:

### Use-case
Currently with version 0.13 the init prompt is the following
```shell
$ reuse init
Initializing project for REUSE.

What license is your project under? Provide the SPDX License Identifier.
To stop adding licenses, hit RETURN.
GPL-3.0-or-later

What other license is your project under? Provide the SPDX License Identifier.
To stop adding licenses, hit RETURN.


What is the name of the project?
myproject   

What is the internet address of the project?
https://....org

What is the name of the maintainer?
Best Maintainer

What is the e-mail address of the maintainer?
thebest@maintainer.org

All done! Initializing now.

Downloading GPL-3.0-or-later
LICENSES/GPL-3.0-or-later.txt already exists

Creating .reuse/dep5

Initialization complete.
```

I think we can make it something like:
```shell
$ reuse init
Initializing project for REUSE.

# A license was found in the pyproject.toml and translated to the SPDX identifier.
What license is your project under? GPL-3.0-or-later was found in pyproject.toml would you like to add it? [Y/n]
 Y

# As this is a multiple-input, the user is prompted for additions
What other license is your project under? Provide the SPDX License Identifier.
To stop adding licenses, hit RETURN.

# Again an existing option is found, this time only a single answer is desired.
What is the name of the project? The name 'default-template' was found in pyproject.toml. Would you like to use it? [Y/n]
 n

# As the suggestion is not used, the original prompt is shown
What is the name of the project?
 myproject   

# No internet address was found, so a regular prompt is shown
What is the internet address of the project?
 https://....org

[...]

PR #70, opened 2019-08-13
Add initial support for templates

### LINKED ISSUE #60: Templates for addheader
We want to support custom templates for headers. Allow default templates per file type and verify integrity.

Some notes that we took:

- [x] Jinja seems like a good templating system for this
- [ ] Add a way to configure default templates per file type
- [x] Verify integrity of template (is a single comment block, contains two tag loops)
- [x] Make sure that the template can replace itself in a header
- [x] [optional] Perhaps allow a method for uncommented templates that can be used for all file types recognised by REUSE? It's fairly trivial to comment such a text block
- [x] Allow a method for commented templates

carmenbianca:
Fixes #60

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ The versions follow [semantic versioning](https://semver.org).
 
 - Added `--year` and `--exclude-year` to `reuse addheader`.
 
+- Added `--template` to `reuse addheader`.
+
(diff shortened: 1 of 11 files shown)


carmenbianca:
This is basically done. I am deferring configuration to #68. Just need to make the linter happy and I can merge.

Issue #18, opened 2019-05-22
Handle deprecated licences and the "+" operator


carmenbianca:
The tool will currently not recognise `GPL-3.0` or `GPL-3.0+` as a valid licence, preferring instead the new `GPL-3.0-only` and `GPL-3.0-or-later`.

Should the tool consider deprecated licences as valid? Should there be a warning?

The "+" operator remains a part of the SPDX spec, however, and should probably be handled correctly.

mxmehl:
If SPDX considers it as valid, let's do the same, but show a warning with a few pointers what to change. This will hopefully trigger devs to update the expression accordingly.

silverhook:
The `+` syntax has not gone away. It is only that (to accommodate for FSF’s position) the GPL family is using `-only`/`-or-later` instead. All the other licenses that have an “or later” option/clause, continue to use the `+` syntax.

ganwell:
I just spend a very frustrating hour, trying identifier after identifier. Internet search tends to find mostly the deprecated identifiers.

Please don't accept these identifiers, but show a message that the identifier is deprecated and maybe a link to a update list.

carmenbianca:
> and maybe a link to a update list.

I can do this in 5 seconds without any effort of actually handling deprecated licenses. Doesn't fix this issue, but would save people time. Thank you for that suggestion!

carmenbianca:
Added messages in 0b6806e

amerlyq:
> All the other licenses that have an “or later” option/clause, continue to use the + syntax.

Does it means I can demand e.g. CC-BY-SA-4.0+ or Apache-2.0+ ? :)
If it only works for specific licenses and SPDX `+` mechanics are not meant for any arbitrary license, then I would appreciate some error message about "this and only this version allowed" being printed instead of assigning generic `Bad license` category. 
It's somewhat similar in spirit with notification about deprecated licenses, is it not?

I spent significant time trying to understand what was wrong with them.

mxmehl:
@carmenbianca, where do we stand here? It seems this is blocking #18

carmenbianca:
It's not really blocking #89. The issue identified in the comments of #89 is a separate issue from #87, which it attempts to fix.

This should probably be tackled at some point, though…

PR #71, opened 2019-08-13
Add some helpful messages referring to the SPDX License List

carmenbianca:
See #18


carmenbianca:
I can only reproduce the linter error locally when I run `pylint -j1 src`. That's a bit strange.

PR #79, opened 2019-08-28
add --include-submodules to lint

### LINKED ISSUE #29: Git submodules should be handled
When I look into https://git.fsfe.org/reuse/website,

this repository has a single license file and a git submodule. The submodule contains a `LICENSES` folder (in master branch).

The first thing I notice is that some files have no SPDX license identifier, for example `README.md`, though the rendered output as 'content of the website' _is_ licensed. 

Files like `.gitmodules` and `Dockerfile` do not have a license. I was hoping to use `reuse lint` in a build step to check whether the project is `REUSE compliant` or not, but this doesn't work since there is no distinction between exceptions and missing license identifier. 

When using submodules I receive the following git error: `fatal: Pathspec 'reuse-docs/.git' is in submodule 'reuse-docs'`

Finally, licenses from submodules cannot be found: `reuse - WARNING - reuse-docs/practices/2.0.md is licensed under CC-BY-SA-4.0, but its license file could not be found`

The reason for this is that the license only exist in the folder of the submodule and is missing in the root folder of the repository. 

In this case, should I copy all licenses from submodules into the root `LICENSES` folder, according to `practice 1` or should I make a reference in the header:

```
/*
* This file is part of project X. It’s copyrighted by the contributors
* recorded in the version control history of the file, available from
* its original location http://git.example.com/X/filename.c
*
* SPDX-License-Identifier: GPL-3.0
*/
```

RipCazza:
fixes #29.

### DIFF
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -247,11 +247,6 @@ def lint_summary(report: ProjectReport, out=sys.stdout) -> None:
 
 def add_arguments(parser):
     """Add arguments to parser."""
-    parser.add_argument(
-        "--include-submodules",
-        action="store_true",
-        help=_("do not include current or specified year in statement"),
-    )
     parser.add_argument("path", action="store", nargs="*", type=PathType("r"))
 
 
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -247,12 +247,18 @@ def lint_summary(report: ProjectReport, out=sys.stdout) -> None:
 
 def add_arguments(parser):
     """Add arguments to parser."""
+    parser.add_argument(
+        "--include-submodules",
+        action="store_true",
+        help=_("do not include current or specified year in statement"),
+    )
     parser.add_argument("path", action="store", nargs="*", type=PathType("r"))
 
 
 def run(args, out=sys.stdout):
     """List all non-compliant files."""
     project = create_project()
+    project.include_submodules = args.include_submodules
     paths = args.path
     if not paths:
         paths = [project.root]
--- a/src/reuse/project.py
+++ b/src/reuse/project.py
@@ -43,7 +43,7 @@ class Project:
     interactions.
     """
 
-    def __init__(self, root: PathLike):
+    def __init__(self, root: PathLike, include_submodules: bool = False):
         self._root = Path(root)
         if not self._root.is_dir():
             raise NotADirectoryError("{} is no valid path".format(self._root))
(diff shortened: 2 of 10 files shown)


src/reuse/lint.py
```diff
@@ -247,12 +247,18 @@ def lint_summary(report: ProjectReport, out=sys.stdout) -> None:
 
 def add_arguments(parser):
     """Add arguments to parser."""
+    parser.add_argument(
+        "--include-submodules",
+        action="store_true",
+        help=_("do not include current or specified year in statement"),
```
carmenbianca: This help text is copied from --year and needs to be changed

src/reuse/project.py
```diff
@@ -92,6 +93,10 @@ def all_files(self, directory: PathLike = None) -> Iterator[Path]:
                 if self._is_path_ignored(root / dir_):
                     _LOGGER.debug("ignoring %s", root / dir_)
                     dirs.remove(dir_)
+                if (
+                    root / dir_ / ".git"
+                ).is_file() and not self.include_submodules:
+                    dirs.remove(dir_)
```
carmenbianca: Would probably be nice to do a `_LOGGING.info()` here

sschuberth:
@RipCazza are you planning to finish this PR up? I'm asking as I'm also interested in this.

And just as a note to myself: The PR title is a bit misleading as it sounds as if this adds linting of submodules as a new feature whereas the opposite is the case: Currently *reuse* unconditionally lints the whole Git repo incl. submodules. This PR makes linting of submodules optional and in fact *disables* it by default (which I like), but allows to enable it again (i.e. restore the current behavior) by passing `--include-submodules`.

tests/conftest.py
```diff
@@ -120,13 +120,45 @@ def git_repository(fake_repository: Path, git_exe: Optional[str]) -> Path:
     build_dir.mkdir()
     (build_dir / "hello.py").touch()
 
-    subprocess.run([GIT_EXE, "init", str(fake_repository)])
-    subprocess.run([GIT_EXE, "add", str(fake_repository)])
-    subprocess.run([GIT_EXE, "commit", "-m", "initial"])
+    subprocess.run([git_exe, "init", str(fake_repository)])
```
sschuberth: Out of curiosity, why is this lower case now?

carmenbianca: I would imagine to use the `git_exe` fixture instead of the global variable. They are practically identical, but using a local variable instead of a global variable is a little bit nicer.

sschuberth:
@carmenbianca could we get a new patch release with this merged?

carmenbianca:
@sschuberth Sure. There were a couple more backwards-incompatible changes I wanted to get in before the next version bump, but they're going to take a little while longer for me.

I'll double-check everything and push 0.6.0 to PyPI.

carmenbianca:
@sschuberth Pushed 0.6.0 to PyPI.

PR #96, opened 2019-10-21
Add NixOS package link to README

davidak:
I'm actually not sure if that is the most accurate representation since the package manager is actually [Nix](https://nixos.org/nix/) and can be used on every linux distro, macOS and in theory also *BSD and Windows.


mxmehl:
Great that reuse is also available there. Thank you!

PR #119, opened 2019-11-14
Implement `--root` argument

### LINKED ISSUE #94: lint fails when called outside of project repo
When running reuse lint not inside the repository but at least one level above, reuse lint does complain about missing files in LICENSES/, although they are in the given directory,

Situation:
```
/home/max/
|─ reuse-example
|─ some_other_dir
```

When running `reuse lint` inside of `/home/max/reuse-example`, everything is fine.

When running `reuse lint reuse-example` inside of `/home/max`, it fails. The same if I run `reuse lint ../reuse-example` when running it inside the `some_other_dir`:

```
# MISSING LICENSES

'CC-BY-4.0' found in:
* reuse-example/img/cat.jpg

'CC0-1.0' found in:
* reuse-example/.gitignore

'GPL-3.0-or-later' found in:
* reuse-example/Makefile
* reuse-example/README.md
* reuse-example/img/dog.jpg
* reuse-example/src/main.c


# SUMMARY

* Bad licenses:
* Missing licenses: CC-BY-4.0, CC0-1.0, GPL-3.0-or-later
* Unused licenses:
* Used licenses:
* Read errors: 0
* Files with copyright information: 6 / 6
* Files with license information: 6 / 6
```

Not sure how a good solution could look like since I assume that it is not trivial for the tool to detect what's the root of the project, especially when submodules are included.

carmenbianca:
Fixes #94 

Closes #100 

This is an alternative approach to #100 that makes the declaration of the root directory explicit.

It also adds a sanity check: The root of a project must be the parent of all paths that are given to the linter. e.g., if the current working directory is the project root, you can't `reuse lint ../`.

This seems like a sufficient "fix"?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,19 @@ following sections:
 
 The versions follow [semantic versioning](https://semver.org).
 
+## Unreleased - YYYY-MM-DD
+
+### Added
+
+- Implemented `--root` argument to specify the root of the project without
+  heuristics.
+
+### Changed
+
+- `lint` no longer accepts path arguments. Where previously one could do `reuse
+  lint SUBDIRECTORY`, this is no longer possible. When linting, you must always
+  lint the entire project. To change the project's root, use `--root`.
+
 ## 0.6.0 - 2019-11-19
 
 ### Added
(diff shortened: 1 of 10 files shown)


carmenbianca:
Also as a small pat on my back, this PR was written totally without internet access :)

carmenbianca:
Last two commits remove the ability to do `reuse lint DIRECTORY`. Instead, `reuse --root=DIRECTORY lint` is used. This is more universally applicable across other subcommands, though I personally find the former syntax far more appealing.

PR #127, opened 2019-11-21
Add multiprocessing

carmenbianca:
Okay so this is a fun one. Ignore all the errors: That's all normal. I turned on multiprocessing for the linting process. Just to see if it could be done, and just to see if it has any perceivable speed increase.

... And it does have a speed increase! I haven't done any professional, thorough, scientific measuring, but here's the result from one run each:

|           Multiprocessing            | Time (mm:ss) |
| :----------------------------------: | :----------: |
| Linux kernel without multiprocessing |     1:15     |
|  Linux kernel with multiprocessing   |     0:39     |
| reuse-tool without multiprocessing | 0:01.01 |
| reuse-tool with multiprocessing | 0:01.13 |
| CPython without multiprocessing | 0:06.00 |
| CPython with multiprocessing | 0:03.19 |

Running on an eight-core Ryzen 3500u.

Really cool stuff! I just have no idea whether it's a good idea. It makes debugging hell, and the speed increase isn't all that noticeable for 99% of projects.

... But it's really cool though!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,14 @@ The versions follow [semantic versioning](https://semver.org).
 - Deprecated licenses are now recognised. `lint` will complain about deprecated
   licenses.
 
+- ProjectReport generation (`lint`, `spdx`) now uses Python multiprocessing,
+  more commonly called multi-threading outside of Python. This has a significant
+  speedup of approximately 300% in testing. Because of overhead, performance
+  increase is not exactly linear.
+
+- For setups where multiprocessing is unsupported or unwanted,
+  `--no-multiprocessing` is added as flag.
+
 ### Changed
 
 - The linter output has been very slightly re-ordered to be more internally
--- a/src/reuse/_main.py
+++ b/src/reuse/_main.py
@@ -71,6 +71,11 @@ def parser() -> argparse.ArgumentParser:
         help=_("do not skip over Git submodules"),
     )
     parser.add_argument(
+        "--no-multiprocessing",
+        action="store_true",
+        help=_("do not use multiprocessing"),
+    )
+    parser.add_argument(
         "--root",
         action="store",
         metavar="PATH",
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -333,7 +333,9 @@ def add_arguments(parser):  # pylint: disable=unused-argument
 def run(args, project: Project, out=sys.stdout):
     """List all non-compliant files."""
     # pylint: disable=unused-argument
-    report = ProjectReport.generate(project, do_checksum=False)
+    report = ProjectReport.generate(
+        project, do_checksum=False, multiprocessing=not args.no_multiprocessing
+    )
     result = lint(report, out=out)
 
     return 0 if result else 1
--- a/src/reuse/report.py
+++ b/src/reuse/report.py
@@ -11,7 +11,7 @@ import random
 from gettext import gettext as _
 from hashlib import md5
 from io import StringIO
-from os import PathLike
+from os import PathLike, cpu_count
 from pathlib import Path
 from typing import Iterable, List, NamedTuple, Optional, Set
 from uuid import uuid4
@@ -180,7 +180,7 @@ class ProjectReport:  # pylint: disable=too-many-instance-attributes
         cls,
         project: Project,
         do_checksum: bool = True,
-        multiprocessing: bool = True,
+        multiprocessing: bool = cpu_count() > 1,
     ) -> "ProjectReport":
         """Generate a ProjectReport from a Project."""
         project_report = cls(do_checksum=do_checksum)
--- a/src/reuse/report.py
+++ b/src/reuse/report.py
@@ -10,9 +10,10 @@ import random
 from gettext import gettext as _
 from hashlib import md5
 from io import StringIO
+from multiprocessing import Pool
 from os import PathLike
 from pathlib import Path
-from typing import Iterable, List, Set
+from typing import Iterable, List, NamedTuple, Optional, Set
 from uuid import uuid4
 
 from . import __version__
@@ -20,6 +21,37 @@ from ._util import _LICENSING, _checksum
 from .project import Project
 
 _LOGGER = logging.getLogger(__name__)
+
+
+class _MultiprocessingContainer:
+    """Container that remembers some data in order to generate a FileReport."""
+
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, project, do_checksum):
+        self.project = project
+        self.do_checksum = do_checksum
+
+    def __call__(self, file_):
+        # pylint: disable=broad-except
+        try:
+            return _MultiprocessingResult(
+                file_,
+                FileReport.generate(
+                    self.project, file_, do_checksum=self.do_checksum
+                ),
+                None,
+            )
+        except Exception as exc:
+            return _MultiprocessingResult(file_, None, exc)
+
+
+class _MultiprocessingResult(NamedTuple):
+    """Result of :class:`MultiprocessingContainer`."""
+
+    path: PathLike
+    report: Optional["FileReport"]
+    error: Optional[Exception]
 
 
 class ProjectReport:  # pylint: disable=too-many-instance-attributes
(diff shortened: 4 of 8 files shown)


carmenbianca:
Wrote a little script here to run some profiling:

https://gist.github.com/carmenbianca/063e1ac06a82490c827c6cf4baf2a5db

Result from master:

master svg

Result from multiprocess:

multiprocess svg

mxmehl:
Oh wow nice!

- Could you please elaborate why it makes debugging so much harder?
- May this have a negative effect when multiple checks are run at the same time, probably each in its own Docker container, e.g. for our API server?

The effect is remarkable for large projects, but as you said neglectable for smaller projects. If it doesn't create too many negative side-effects, it's very nice though!

carmenbianca:
> Could you please elaborate why it makes debugging so much harder?

The code spawns a separate child process for every single file that is being linted. I did this under the assumption that parsing the contents of a file (with regexes) is the most resource-intensive part of the linter. I also did it because it's the easiest thing to multi-process.

But a debugger can't or doesn't attach itself to the child process. So if something goes wrong in the child process, it's like a black box that you can't really peek into it. Or at least, that is my understanding. Maybe it's possible, but it's not the easiest thing in the world.

Fortunately, there are two ways to circumvent that problem:

- Test the stuff that happens in the child process without any multiprocessing. *This is already in the test suite*, so that's easy.
- Add a switch for multiprocessing so that it can be turned off/on. This is fairly trivial, but adds a non-zero chance that the result differs between the two paths, because they have at least a little bit of code that is different between them. I doubt that's a big problem, though.

> May this have a negative effect when multiple checks are run at the same time, probably each in its own Docker container, e.g. for our API server?

I think the server only has a single CPU thread, so the program only ever spawns a single child process at a time. If so, there's a very very tiny speed decrease due to some additional overhead. Otherwise, nothing has changed.

> The effect is remarkable for large projects, but as you said neglectable for smaller projects. If it doesn't create too many negative side-effects, it's very nice though!

Thanks :-) It's not exactly production-ready yet, though. I hacked this up quite quickly, and it 99% works, but the last 1% is a pain. If the linter can't read a certain file (e.g., it doesn't have the correct permissions), the whole thing collapses, which isn't ideal.

Unfortunately I can't work out how to fix that. For my own reference because I'm going to stop writing code after writing this comment:

- `error_callback` in `(star)map_async` *works* in that it executes a bit of code when an error is encountered, but the exception is still raised when you do `result.get()`. This is really really really annoying.

- Probably need to handle the errors inside of the workers, and return a common data type or something?

- Alternatively, use `apply_async` and it's *probably* possible to handle each error separately. However, I tried doing `apply_async` (without any error handling), and it's *much slower*.

- The above slowdown is probably because `project.all_files` is fairly slow? The `map` functions simply run `list(project.all_files)` in the main process at the start. Which somehow strangely makes the whole thing faster.

carmenbianca:
> The code spawns a separate child process for every single file that is being linted

Correction: The code spawns an amount of processes equal to the amount of threads, and all the files are distributed across these processes. Like separate queues in front of cash registers in a shop.

carmenbianca:
> Probably need to handle the errors inside of the workers, and return a common data type or something?

This is what I ended up doing. It works fairly well.

> Alternatively, use apply_async and it's probably possible to handle each error separately. However, I tried doing apply_async (without any error handling), and it's much slower.

Still much slower. I tried computing `project.all_files()` in advance, and it doesn't change anything. I'm frankly not sure why this is.

---

Anyway, all the tests are running correctly now, and there are no glaring errors that I can see. I'm wondering if this should have a switch (e.g. `reuse --no-multiprocessing lint`)? I'm not entirely sure who would want to use that switch, however.

Also, ideally, I'd somehow optimise `project.all_files`, because it's really really really slow. This is mostly the fault of `pathlib`, I think. I'm not really sure.

mxmehl:
Nice, thanks!

> Anyway, all the tests are running correctly now, and there are no glaring errors that I can see. I'm wondering if this should have a switch (e.g. reuse --no-multiprocessing lint)? I'm not entirely sure who would want to use that switch, however.

Hm, if it's not too much of a hassle to implement this, why not? If you need to debug things or if exotic python versions or OSs are being used, it might make a positive difference, or?

carmenbianca:
> Hm, if it's not too much of a hassle to implement this, why not?

I'll take a look at this tomorrow. Getting some brain aneurysms from trying to optimise `project.all_files`. I suspect that adding a switch won't be very difficult.

carmenbianca:
Writing this for posterity's sake again:

It's possible to take out all multiprocessing with a single line of code. That is, replace:

    pool = multiprocessing.Pool()

with

    pool = multiprocessing.dummy.Pool()

This is a wrapper around the `threading` module. I tried it, and it works. Problems are, however:

- This still makes debugging difficult.
- There's a 10% performance penalty compared to the master branch (i.e., fully single-threaded).

So maybe not the best solution.

carmenbianca:
Okay so this is ridiculous. I have found an additional speedup on the Linux kernel from 30 seconds to 20 seconds. Patch at the end of this comment.

It is the most stupid patch, however. `project.all_files` is slow because it checks, for every individual file, whether that file is ignored by Git or not. In itself, this is just checking whether a file exists inside of a set, which is a really quick operation. However, the list of files is relative to the root of the project, while the result of `project.all_files` is relative to is relative to the root of the project *as relative to the current working directory*.

For instance, if my CWD is `~/`, and I do `reuse lint Projects/reuse-tool`, then I might get `Projects/reuse-tool/Makefile` as one of the yielded files. However, in order to compare that path to the list of ignored files, `Projects/reuse-tool` must be culled from that path. This happens here:

```python
    def _relative_from_root(self, path: PathLike) -> Path:
        """If the project root is /tmp/project, and *path* is
        /tmp/project/src/file, then return src/file.
        """
        return Path(os.path.relpath(path, start=self.root))
```

It turns out, if you do this many times, this is computationally intensive. The patch I have written simply takes that single operation, and distributes it across the cores for all files. This is stupid and pointless and ugly.

Ideally, I would like to make that operation unnecessary. But I don't know how to do that. Here are some ideas:

- Have `project.all_files` yield files relative only to the root of the project. This means that, if you want to do anything with those files, you must afterwards prepend the path to the root. Maybe this helps? I don't know.

- The list of ignored files should be relative to the project root as relative to the CWD instead. This is really stupid, and possibly slower than the current method if the list of ignored paths is large.

- Somehow change directory to the root of the project inside of `project.all_files`. Not impossible, but a bit weird?

I'm a bit lost there. This patch, however, does away with any actual thinking and just brute-forces the problem:

```patch
diff --git a/src/reuse/project.py b/src/reuse/project.py
index ead2ed0..4bf8430 100644
--- a/src/reuse/project.py
+++ b/src/reuse/project.py
@@ -10,6 +10,7 @@ import os
 from gettext import gettext as _
 from pathlib import Path
 from typing import Dict, Iterator, Optional
+from multiprocessing import Pool
 
 from boolean.boolean import ParseError
 from debian.copyright import Copyright
@@ -39,6 +40,22 @@ from ._util import (
 _LOGGER = logging.getLogger(__name__)
 
 
+class _MultiprocessingContainer:
+    """Container that remembers some data in order to generate a FileReport."""
+
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, project):
+        self.project = project
+
+    def __call__(self, file_):
+        # pylint: disable=broad-except
+        if not self.project._is_path_ignored(file_):
+            return file_
+        else:
+            return None
+
+
 class Project:
     """Simple object that holds the project's root, which is necessary for many
     interactions.
@@ -83,6 +100,8 @@ class Project:
             directory = self.root
         directory = Path(directory)
 
+        all_files = []
+
         for root, dirs, files in os.walk(directory):
             root = Path(root)
             _LOGGER.debug("currently walking in '%s'", root)
@@ -104,12 +123,26 @@ class Project:
             # Filter files.
             for file_ in files:
                 the_file = root / file_
-                if self._is_path_ignored(the_file):
-                    _LOGGER.debug("ignoring '%s'", the_file)
-                    continue
+                # if self._is_path_ignored(the_file):
+                #     _LOGGER.debug("ignoring '%s'", the_file)
+                #     continue
 
                 _LOGGER.debug("yielding '%s'", the_file)
-                yield the_file
+                all_files.append(the_file)
+
+        pool = Pool()
+
+        container = _MultiprocessingContainer(self)
+
+        results = pool.map(container, all_files)
+
+        pool.close()
+        pool.join()
+
+        for result in results:
+            if result:
+                yield result
+
 
     def spdx_info_of(self, path: PathLike) -> SpdxInfo:
         """Return SPDX info of *path*.
```

carmenbianca:
This bit of 8459b61 improved performance quite significantly:

```patch
-    def _relative_from_root(self, path: PathLike) -> Path:
+    def _relative_from_root(self, path: Path) -> Path:
         """If the project root is /tmp/project, and *path* is
         /tmp/project/src/file, then return src/file.
         """
-        return Path(os.path.relpath(path, start=self.root))
+        return path.relative_to(self.root)
```

It breaks the tests, but otherwise performs a lot better. It gets the time down to 22 seconds from 30. If paired with the above ridiculous patch, it would probably perform even better.

carmenbianca:
6002ff4 added a fallback to the old method. The fallback really shouldn't be necessary, but it's there.

carmenbianca:
Pushed some improvements back to master. master now has a speed of 0:55 compared to 0:22 in this branch. General improvements round the board.

carmenbianca:
> > Anyway, all the tests are running correctly now, and there are no glaring errors that I can see. I'm wondering if this should have a switch (e.g. reuse --no-multiprocessing lint)? I'm not entirely sure who would want to use that switch, however.
> 
> Hm, if it's not too much of a hassle to implement this, why not? If you need to debug things or if exotic python versions or OSs are being used, it might make a positive difference, or?

Implemented this. This PR is looking pretty mature right now.

carmenbianca:
Note to self: Disable multiprocessing if the host only has a single core/thread. Bit of a no-brainer really.

PR #133, opened 2019-11-30
Create a GitHub Actions workflow

carmenbianca:


carmenbianca:
Today I learnt that reuse is incompatible with Windows. Or rather, if reuse on Windows encounters a file with Unix newlines, it fails.

Why, Windows, why? Why can't you just be easy?

carmenbianca:
Oof that was a labour and a half. Still need to make sure that the dos files in fake_repository/src are actually CRLF after git checkout, but at least it works now.

The problem was: When a binary file containing CRLF line endings is converted to a Unicode string, it keeps the `\r\n` line endings. The `re` module does not recognise those line endings at all. I have no idea why.

PR #134, opened 2019-11-30
Add CoC

carmenbianca:
Should probably add a reference in the README.

There's also a small problem: I do not know the license of the CoC.


carmenbianca:
Maybe a better idea: Have `CODE_OF_CONDUCT.md` be a relatively small file that contains a link to https://fsfe.org/about/codeofconduct instead of containing the verbatim text. This makes sure that upstream changes are immediately applied without any shred of a doubt.

mxmehl:
Yes, linking it makes sense! You can consider the text to be under CC-BY-SA-4.0, author being the FSFE

Issue #130, opened 2019-11-29
Docker latest points to master branch


carmenbianca:
See #128 

@mxmehl

mxmehl:
Well, `latest` has to point to a specific tag or branch. What we could to is having the `master` branch holiding experimental stuff which creates the `dev` Docker image, and a `stable` branch or tag which resolves to the latest release that we consider to be stable.

This solution would allow for automated build. Otherwise, I would have to build each tag manually which is rather painful and error-prone. @carmenbianca, what do you think about a stable branch or moving tag?

carmenbianca:
My understanding of Docker conventions is that `latest` points to the latest stable release, like an alias. What you're proposing would be the `stable` branch. I don't know the DockerHub admin UI at all, however. Is there no button that basically just "does that"?

`dev` or `master` could point to the master branch, sure.

mxmehl:
> My understanding of Docker conventions is that `latest` points to the latest stable release, like an alias. What you're proposing would be the `stable` branch. I don't know the DockerHub admin UI at all, however. Is there no button that basically just "does that"?

Unfortunately there is no option to make `latest` a pointer to the last tagged release. I need either a tag or branch to refer to, unless I make everything manually, and that's not really my goal to be honest...

carmenbianca:
OK, I created a DockerHub account to see what the UI looks like. Would it help if I created a `latest` branch on GitHub that always points to the latest tagged release? It's a little bit more busywork, but a lot less busywork than messing around with the DockerHub admin UI every release.

mxmehl:
Yes, that's possible! For Docker Hub it does not seem to matter whether it's a branch or tag, so feel free to choose :)

carmenbianca:
Ah, a `latest` tag would be even easier.

I'll create the `latest` tag, and we'll end up with this:

latest tag (GitHub) -> latest (Docker)
master branch (GitHub) -> dev (Docker)

mxmehl:
Ah thanks. I've set the auto-build rules like this. Is there any way to verify that we are on the correct commit and/or version, i.e. to find out whether `latest` is on 0.7.0 or master?

carmenbianca:
> Is there any way to verify that we are on the correct commit and/or version, i.e. to find out whether latest is on 0.7.0 or master?

[...]

PR #142, opened 2019-12-05
Implement setuptools-scm

carmenbianca:
@mxmehl This patch implements the behaviour you mentioned in #130, with fallback behaviour.

I can't formally write tests for this, so I'll do some additional playing around before I merge.


carmenbianca:
Done sufficient testing, merging this.

mxmehl:
Thanks! The additional dependency is probably available on all operating systems which is good for packaging :)

PR #152, opened 2020-01-24
Suggest valid SPDX License Identifiers when invalid one is given

siiptuo:
Suggest corrections for typos like "GLP3-or-lter" and unclear licenses like "BSD" in init and download commands.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -18,8 +18,6 @@ Contributors
 
 - Sebastian Schuberth <schuberth@fsfe.org>
 
-- Kirill Elagin
-
 - Max Mehl <max.mehl@fsfe.org>
 
 - Matija Šuklje <hook@fsfe.org>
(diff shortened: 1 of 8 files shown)


carmenbianca:
Hi @siiptuo! Thank you for the pull request.

Unfortunately I won't have time to review it in the coming week/two weeks. After that I'll do my best to review it.

carmenbianca:
I've replaced the jaro dependency with a function from the standard library. I did this because the jaro library is not available in the repositories of some target Linux distributions.

Will do a little more tinkering soon to decouple the logic of generating the error message from the logic of finding similar licenses.

carmenbianca:
Did all the tinkering. Encountered an issue with attempting to download deprecated licenses, though, so will create a separate issue for that.

siiptuo:
Your changes look great! I made a small final fix and rebased everything.

carmenbianca:
Thanks @siiptuo and sorry for the super long wait on this PR

PR #154, opened 2020-02-04
allow tab after SPDX-License-Identifier

pH5:
The SPDX Specification does not specify that the whitespace between the `SPDX-License-Identifier:` string and the license expression must be a single space. The regular expression already matched multiple spaces or space-and-tab combinations before, so allowing a tab directly after the `SPDX-License-Identifier:` string improves consistency.


carmenbianca:
LGTM

PR #167, opened 2020-02-09
Improve Dockerfile

carmenbianca:
Tinkered a bit with the Dockerfile. This was challenging because `COPY . /reuse-tool/` somehow takes ages on my system.

I split out the commands and added a `CMD` at the end, which is a default command if no custom command is supplied.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Install reuse
       run: |
         python setup.py install
@@ -44,7 +44,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Install reuse
       run: |
         python setup.py install
@@ -63,7 +63,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Lint with PyLint
       run: |
         make lint
@@ -79,7 +79,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Test formatting with black
       run: |
         make blackcheck
@@ -95,7 +95,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Install reuse
       run: |
         python setup.py install
@@ -114,7 +114,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-dev.txt
     - name: Install reuse
       run: |
         python setup.py install
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 # Create a base image that has git installed.
-FROM python:3.8-slim AS base
+FROM python:3.8-alpine AS base
 
-RUN apt-get update
-RUN apt-get install -y --no-install-recommends git
-
+RUN apk --no-cache add git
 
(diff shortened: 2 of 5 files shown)


mxmehl:
Thanks! Could you please explain why you changed the cd/install/rm flow in your approach? If you run the rm in a separate step, it doesn't make any difference in image size as the layer will already be there.

Also, what is the idea behind having the lint as default CMD? To save a separate call for people using it in CI?

carmenbianca:
> it doesn't make any difference in image size as the layer will already be there.

I wasn't aware of this at all. So armed with this knowledge, I went ahead and redid the whole thing in 
bd763a0, using [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/).

Basically, I'm using two images to build a third image. The first image is the "base" image, running `python:3.8-alpine` with Git installed. The second image uses the base image to "compile" everything, as it were, into a virtualenv. The third image uses the base image and copies the virtualenv over.

Using this approach, all of the artifacts from the second image are discarded (or rather, not included in the third image). Artifacts such as the Git repo of reuse-tool, the cache of downloaded files from PyPI, general build artifacts, and so forth.

I also froze the requirements against requirements.txt, so that the output is more deterministic. It's not 100% deterministic though, because second-level requirements are not frozen. As part of doing this, I split off the runtime requirements from the dev requirements.

… Even so, the resulting image is bigger. Give-or-take 80 MB vs 140 MB. The reason for that is that I'm using `python:3.8-alpine` instead of plain `alpine:latest`. The reason for that choice is also determinism. I could reduce the size by using plain `alpine:latest`, but I'm not sure if that's a good trade-off or not.

> Also, what is the idea behind having the lint as default CMD? To save a separate call for people using it in CI?

Yep. On the CLI, you can do `docker run -v $(pwd):"/data" -it <reuse-image>` to run the linter on the current working directory. For GitHub Actions, this would suffice:

```
steps:
...
- name: Check REUSE compliance
        uses: docker://fsfe/reuse
```

`reuse lint` is not required. You can still type it, of course, and it would simply override CMD. But it's a cool nifty little thing.

mxmehl:
Thanks for the long explanation. Multi-staged builds were new to me as well, but they make sense!

I compared the sizes because I would like to keep the image size as small as possible. All sizes compressed in MB:

* Current: 29.3
* This PR: 46.5
* This PR, but using alpine:latest: 31.5

I cannot tell how important a more deterministic behaviour is for this project, so I would trust your assessment whether or which way to go here.

Mental note: if we change the Dockerfile here, check the instructions on reuse.software/dev and the compatibility with the current API's workflow.

carmenbianca:
> I cannot tell how important a more deterministic behaviour is for this project

Basically, when Python 3.9 is released, and `alpine:latest` starts installing Python 3.9 instead of 3.8 with its package manager, there is a risk that the image will stop working (and there are no tests in place that will signal us when/if it stops working). Python releases are *not* strictly backwards compatible, which is kind of annoying.

… However, by the time 3.9 is released, and by the time Alpine includes it in their latest release, we will probably have support for 3.9 already, so it might be a moot point.

Alternatively, instead of using `alpine:latest`, we could use `alpine:3.11`. We would simply need to bump that once every 6-12 months. This gives both determinism and small image size at the cost of putting a frequent reminder somewhere, somehow. Probably the best solution?

I'm not actually any good at Docker tbh. I simply spent the weekend tinkering on the Dockerfile because I ran into [this blog](https://pythonspeed.com/docker/). That blog recommends using Debian instead of Alpine because of [these reasons](https://pythonspeed.com/articles/alpine-docker-python/), but to the best of my knowledge, these corner cases don't apply because none of reuse-tool's dependencies have compiled C code.

mxmehl:
Hm, I see your point, and it's a hard decision.

Could you think of an automated test which builds the Docker image and runs one command? This could allow us to use xy:latest and see rather quickly whether something broke.

On the other hand, we could bump the base image version every time we make a major release, so use this occasion as a reminder.

> I'm not actually any good at Docker tbh. I simply spent the weekend tinkering on the Dockerfile because I ran into [this blog](https://pythonspeed.com/docker/). That blog recommends using Debian instead of Alpine because of [these reasons](https://pythonspeed.com/articles/alpine-docker-python/), but to the best of my knowledge, these corner cases don't apply because none of reuse-tool's dependencies have compiled C code.

The image you use now also makes use of alpine and musl libc instead of glibc, but I also agree that it doesn't really have an effect on reuse-tool.

carmenbianca:
> Could you think of an automated test which builds the Docker image and runs one command

https://hub.docker.com has some kind of Autotest functionality. I don't know how to use it, though. I'll open a new issue for that.

In any case, I switched over to `alpine:3.11`. We'll see how well it works.

PR #172, opened 2020-02-11
Tests for Dockerfile and Docker image

### LINKED ISSUE #171: Enable Autotests for the Docker image
See #167 for some background.

mxmehl:
This PR adds a step in the CI to test whether the Dockerfile can be built and whether the result successfully lints the source of this programme.

Thereby, we should be able to detect whether changes in some base image will result in errors.

Fixes #171

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -121,3 +121,15 @@ jobs:
     - name: Create docs with Sphinx
       run: |
         make docs
+
+  docker:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build Dockerfile
+      run: |
+        docker build -t reuse .
+    - name: Run Docker image
+      run: |
+        docker run -v "$(pwd):/data" reuse


mxmehl:
@carmenbianca any thoughts on the PR?

carmenbianca:
Sorry, hadn't had the time to review for a while. I think this could work. I'm just hesitant about simply calling `reuse` to test if everything is order.

Given that you pass `-v "$(pwd):/data"`, would it be possible to run `pytest` (or `make test`) instead? We'd need to somehow install pytest in the image, but it would run an actual test, which would be great.

mxmehl:
It's calling `reuse lint` against the source at the moment as this is the CMD in the Dockerfile. I would consider this to be fine as a first solution as this is the main subcommand the Docker image is supposed to run.

However, we can also run pytest inside. I'm not sure what the best way would be, but probably a separate Dockerfile instead of a queue of commands to be run inside the image.

carmenbianca:
Ah, I was mistakenly under the impression that it ran `reuse`, which is equivalent to `reuse --help`. I agree that this should be sufficient. Merging.

PR #192, opened 2020-02-29
add file extensions for Isabelle prover

lsf37:
This pull request adds file-extension recognition for 2 types of files used by the [Isabelle theorem prover](http://isabelle.in.tum.de), and one filename match for the same, mapping all three to the existing ml-comment style.

All three are recognised by syntax highlighting/linguist on github and satisfy their uniqueness requirements, so should be safe to add.

(We're in the process of converting a few larger repositories such as https://github.com/sel4/l4v to SPDX tags, and `reuse addheader` recognising these file types would be really handy).


carmenbianca:
@lsf37 Thanks a lot for the PR! Sorry for the wait.

lsf37:
👍 No worries!

PR #196, opened 2020-03-29
Add CMakeLists.txt to filename comment style map

nicolas17:
CMake uses # for comments (`PythonCommentStyle`), and it's already in the extension map as *.cmake -> `PythonCommentStyle`, but `CMakeLists.txt` was missing from the filename mapping.


carmenbianca:
@nicolas17 Thanks a lot for the PR. Merging.

PR #213, opened 2020-04-21
Add pre-commit hook

### LINKED ISSUE #212: Feature Request: run as pre-commit integration
[pre-commit](https://pre-commit.com/) would be a very good way to make sure that all files (particularly, new files) in the repository are given a license. In particular it would reduce the risk that an explicit `.license` file is created locally (`reuse lint` passes) but not added to the right commit.

carmenbianca:
Fixes #212.

Great suggestion by @Flameeyes. Implementing this was super easy. This should also be added to the documentation on https://reuse.software/dev/

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- id: reuse
+  name: reuse
+  entry: reuse lint
+  language: python
+  pass_filenames: false
+  description: "Lint the project directory for compliance with the REUSE Specification"
+  language_version: python3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ The versions follow [semantic versioning](https://semver.org).
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


mxmehl:
So this is not really the classical pre-commit hook, right?

Normally, it would suffice to create a file in ones `.git/hooks/` directory:

```
#!/bin/sh
reuse lint
```

Flameeyes:
It's the integration for pre-commit.com — which is basically a way to manage "repository side" the configuration of hooks. `pre-commit install` will set up the `.git/hooks/pre-commit` file to run one or more pre-commit checks (and in case, a way to fix them). Tools like black, isort, flake8 and mypy integrate with it as well, so it's easy for a repository to add those without the users needing to do anything but the first install.

carmenbianca:
@mxmehl That would also works, but pre-commit does the following things in addition:

- Automatically downloads the latest stable version of reuse for you
- Installs reuse in a virtual environment for you so you don't need to do anything installation-wise

The advantage of doing that is that the tool handles everything for you without any manual intervention. In turn, that gives you some certainty that all developers use the same pre-commit hook.

As a matter of fact, this repo already used pre-commit with [black](https://github.com/psf/black/) to lint formatting :)

mxmehl:
Thanks for the explanation, sounds interesting and makes a lot of sense :)

PR #216, opened 2020-04-21
Add contact address to all FSFE copyright statements

carmenbianca:
This PR is kind of annoying, but it has bothered me for a while that this repo doesn't fully follow the REUSE suggestion to include a contact address in copyright statements.


carmenbianca:
Thanks @mxmehl!

PR #217, opened 2020-04-21
Add --single-line and --multi-line flags to addheader

### LINKED ISSUE #209: addheader --style c is more C++ style (or C99)
This is minor for most people _but_ some projects still support per-C99 compilers that don't allow `//` comments (or that warn on them).

I would suggest renaming the style to `c++` at the very least.

carmenbianca:
Fixes #209.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@ The versions follow [semantic versioning](https://semver.org).
   + XSL (`.xsl`)
   + Mailmap (`.mailmap`)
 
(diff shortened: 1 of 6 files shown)


carmenbianca:
This PR changes a translateable string, which means asking volunteers to update it again.

Would it be easier to move translation to weblate, so that it can be done more continuously?

mxmehl:
Sure, we can do that. I've never set up a weblate project page though.

carmenbianca:
I petitioned for a project on Weblate. Will take a while, and I'll probably need to spend a few hours fiddling with Weblate once it's all set up.

Issue #224, opened 2020-05-01
New addheader option to skip files with unrecognised comment styles


davidag:
As of `reuse 0.10.0` the `addheader` command aborts processing if the comment style of any file can't be autodetected and neither `--style` nor `--explicit` is used.

This behavior is clearly stated on `reuse addheader --help`:

> The comment style should be auto-detected for your files. If a comment style could not be detected, the process aborts. Use --style to specify or override the comment style to use.

I'm applying `reuse` to a repository with many files and my first approach has been this:

    $ git ls-files | xargs reuse addheaders ... 

but due to the presence of unrecognised comment styles, this fails. If I still want to use `reuse`, I believe I'd have to check which files are supported by `reuse` and then build the command to run specifying each of them as arguments. Of course, unrecognised files should be manually changed (I can list those using `reuse lint`).

That's the reason I'd like to propose the possibility of `reuse` not aborting if some files have unrecognised styles, but others can still be processed. This could be achieved with a new `addheader` option. A name for this new option could be, for example, `--skip-unrecognised-files` or `--skip-unrecognised`.

An example `reuse` output could be as follows:

```
$ reuse addheader [required-options] --skip-unrecognised Makefile .madeupfile
Successfully changed header of Makefile
Skipped unrecognised file .madeupfile
```

If this feature is of any interest, I could work on it :relaxed:

carmenbianca:
This sounds like a great feature! If you want to work on it, feel free. It shouldn't be super difficult to implement, I don't think.

carmenbianca:
Fixed by #225

PR #225, opened 2020-05-02
Skip unrecognised files using --skip-unrecognised

davidag:
Implements #224 

I've left all my local commits for the sake of clarity, just let me know if I should squash them into one before merging :+1:


davidag:
Should I extract all `addheader` tests to a new `test_main_addheader.py` file to avoid the linting error or should I disable it?

carmenbianca:
@davidag The PR looks excellent, thank you! I've added a few minor tweaks. I'll merge when the CI has finished running all the tests.

Thanks! :tada:

Issue #221, opened 2020-04-28
Files are read assuming utf-8 but written in locale-dependent encoding


Flameeyes:
Filing this so I don't forget details, but I have no idea exactly how to reproduce this enough to have this work correctly.

When using the reuse tool on Windows via Visual Studio Code/Power Shell, the command line arguments appear to be received as latin-1, and copied directly into the generated file as they are, despite that fact that Python files should be utf-8 if not otherwise stated.

Part of this is clearly a PowerShell issue that reuse can't fix, but at least there should be a warning if the input parameter is not valid utf-8?

```
(venv) PS C:\Users\Diego Elio\Documents\GitHub\pdfrename> python -m hexdump .\tools\pdfrename.py
00000000: 23 20 53 50 44 58 2D 46  69 6C 65 43 6F 70 79 72  # SPDX-FileCopyr
00000010: 69 67 68 74 54 65 78 74  3A 20 32 30 32 30 20 44  ightText: 2020 D
00000020: 69 65 67 6F 20 45 6C 69  6F 20 50 65 74 74 65 6E  iego Elio Petten
00000030: F2 0D 0A 23 0D 0A 23 20  53 50 44 58 2D 4C 69 63  ...#..# SPDX-Lic
```

The `F2` at `0x30` is indeed `ò` in latin-1.

carmenbianca:
Oh dear…

I have no idea how to test this. Checking for valid UTF-8 is possible, probably, but won't catch all encoding errors.

A naive solution would be to detect Powershell and convert to UTF-8. But knowing Microsoft Windows, the input encoding is *probably* dependent on the user's locale settings. e.g., a Japanese version of Windows probably doesn't use Latin-1 in the shell.

This probably points to a slightly more "problematic" thing in the codebase, where UTF-8 is always assumed. One older version of reuse used chardet to detect encoding. Specifically, it was also used to detect binary files and skip them. Ironically, I had implemented that to speed things along (if binary files are skipped, surely that speeds things up). Turns out it didn't, and it's important to measure things before prematurely optimising them.

Instead, the program currently reads all files as binary input and runs them through this function:

```python
def decoded_text_from_binary(binary_file: BinaryIO, size: int = None) -> str:
    """Given a binary file object, detect its encoding and return its contents
    as a decoded string. Do not throw any errors if the encoding contains
    errors:  Just replace the false characters.

    If *size* is specified, only read so many bytes.
    """
    rawdata = binary_file.read(size)
[...]

PR #231, opened 2020-05-15
Always explicitly encode output files in UTF-8

Flameeyes:
This addresses #221 — Python already takes care of decoding arguments correctly, but the default mode of the files depends on `locale.getpreferredencoding(False)` (see https://docs.python.org/3/library/functions.html#open).

Since most languages are now enforcing utf-8 encoding for source code, this should be a positive change for all of them.


Flameeyes:
I have messed things up and created a duplicate pull request, sorry :(

PR #234, opened 2020-05-23
Add go.mod support

ethulhu:
A `go.mod` file is a Go module definition, similar to a makefile.

The other half of `go.mod` is a checksum file `go.sum`, which is just a list of checksums, and is arguably a build artifact, but one that should be checked in to source control, and thus can't be put in `.gitignore`.

### DIFF
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -52,6 +52,9 @@ _IGNORE_FILE_PATTERNS = [
     re.compile(r"^\.gitkeep$"),
     re.compile(r".*\.license$"),
     re.compile(r".*\.spdx$"),
+
+    # go.sum is a machine-maintained list of checksums used by Go modules.
+    re.compile(r".*go.sum$"),
 ]
 
 #: Simple structure for holding SPDX information.
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -52,7 +52,6 @@ _IGNORE_FILE_PATTERNS = [
     re.compile(r"^\.gitkeep$"),
     re.compile(r".*\.license$"),
     re.compile(r".*\.spdx$"),
-
     # go.sum is a machine-maintained list of checksums used by Go modules.
     re.compile(r".*go.sum$"),
 ]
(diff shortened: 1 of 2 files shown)


carmenbianca:
Hi @ethulhu thanks for the submission!

@silverhook Do you have any idea on whether reuse should take a stance towards generated lockfiles? This PR suggests to ignore such files automatically. I found an example of such files here: https://github.com/tresnadery/test_evermos/commit/f874ff8557fce8efefea52d4c3fd7cb55f4957a3

silverhook:
I can’t see how a (simple generated) lockfile (or hash file) would be copyrightable. 

I’m split between saying it should be ignored and it should be just declared as uncopyrightable (e.g. licensed under CC0-1.0).

The problem I see with ignoring it, we’d be opening up a pandora’s box of having to handle all the different files that are (most likely) not copyrightable in the spec and tool, instead of letting the author decide what it is.

While zero-content files are a pretty safe bet that they are not copyrightable, with files with actual (even generated) content, I am worried about us prejudicing whether these are or not capable of being protected by copyright. That decision I think is outside the scope of a spec.

I think it best we poll other lawyers on this topic. SPDX or LN lists would be a good start.

carmenbianca:
> I think it best we poll other lawyers on this topic. SPDX or LN lists would be a good start.

If that is your conclusion then I'd sooner say that the decision should be delegated to the user, instead of determined upstream here. I'll write a message to poll opinions later, though.

mxmehl:
I agree, let's not overcomplicate things by trying to define what's copyrightable (and thereby not ignorable) or not. I agree with @silverhook that zero-content files are a good red line for REUSE.

ethulhu:
Should I split this into 2 PRs, one for auto-detecting `go.mod` and another for discussing `go.sum`?

geistesk:
As an outsider, I would like to see this PR merged.

Especially the automatically generated `go.sum` file will be overwritten when being read. Thus, each time one builds a Golang application, the copyright header will be removed. And this is very tedious.

carmenbianca:
Closed by #243

Issue #13, opened 2019-05-21
Behaviour of `reuse addheader`


carmenbianca:
This issue outlines all the ways in which one might use `reuse addheader`, and what should happen in those cases.

# 1. Simplest case

This is a super straightforward case.

`reuse addheader --copyright "Mary Sue" --license 0BSD myfile.py` should add the header

```
# SPDX-Copyright: CURRENT_YEAR Mary Sue
#
# SPDX-License-Identifier: 0BSD
```

Should there be a confirmation prompt here?

# 2. Overwriting current year

Also a simple case.

`reuse addheader --year 1984 --copyright "Mary Sue" --llicense 0BSD myfile.py` should add the header

```
# SPDX-Copyright: 1984 Mary Sue
#
# SPDX-License-Identifier: 0BSD
```

# 3. Chaining copyright holders and licenses

You can repeat arguments.

`reuse addheader --copyright "Mary Sue" --copyright "John Doe" --license 0BSD --license MIT` should add the header

```
# SPDX-Copyright: CURRENT_YEAR John Doe
# SPDX-Copyright: CURRENT_YEAR Mary Sue
#
# SPDX-License-Identifier: 0BSD
# SPDX-License-Identifier: MIT
```

# 4. No arguments

The thing prompts you, maybe? This might be super clumsy though.

Everything between brackets is user input.

```
$ reuse addheader myfile.py
Who is the copyright holder?: [Mary Sue]
What is the license?: [0BSD]
SUCCESS MESSAGE HERE
```

# 5. Environment variables

Maybe we can set environment variables so that you don't have to be prompted. Kind of like a default setting for lazy users.

```
$ export NAME="Mary Sue"
$ export EMAIL="mary@example.com"
$ export DEFAULT_LICENSE="0BSD"
$ reuse addheader myfile.py
```

will add the header

```
# SPDX-Copyright: CURRENT_YEAR Mary Sue <mary@example.com>
#
# SPDX-License-Identifier: 0BSD
```

Should the user be prompted to confirm this?

# 6. Config

Maybe we could put some defaults in `.reuse/config`. Let's say that that file looks like this:

```
[reuse]
default_license="0BSD"
```

`reuse addheader --copyright "Mary Sue" myfile.py` should add the header

```
# SPDX-Copyright: CURRENT_YEAR Mary Sue
#
# SPDX-License-Identifier: 0BSD
```

Should the user be prompted to confirm this?

[...]

PR #240, opened 2020-07-12
Read default copyright from git/hg configuration

siiptuo:
Use the same configuration files and environment variables as git and hg.

Related to #68

### DIFF
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -1,7 +1,8 @@
 # SPDX-FileCopyrightText: 2019 Free Software Foundation Europe e.V. <https://fsfe.org>
-# SPDX-FileCopyrightText: 2019 Stefan Bakker <s.bakker777@gmail.com>
 # SPDX-FileCopyrightText: 2019 Kirill Elagin <kirelagin@gmail.com>
+# SPDX-FileCopyrightText: 2019 Stefan Bakker <s.bakker777@gmail.com>
 # SPDX-FileCopyrightText: 2020 Dmitry Bogatov
+# SPDX-FileCopyrightText: 2020 Tuomas Siipola <tuomas@zpl.fi>
 # SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
 # SPDX-FileCopyrightText: 2021 Alvar Penning
 #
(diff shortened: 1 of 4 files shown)


mxmehl:
Thank you @siiptuo!

@carmenbianca Since you know that I am a terrible code, would you have time to check this PR?

As I get from the code, it reads the author env from the Git config. We would be careful with making this the default though as `author != copyright holder` in some cases, and especially `REUSE user != copyright holder` in many more cases.

nicorikken:
@siiptuo thanks for this contribution. How does this deal with the mis-usage when the variables exist but are empty?

Personally I would like a small log-statement that the environment-variables or git config is used, because this behavior is quite 'magical'. Perhaps something in the debug mode.

I think this is nice for individual copyright owners and contributors.
As @mxmehl mentions, I don't see this work well in a more corporate environment where a company or organization is the copyright holder. Perhaps some additional environment variables could help with that use-case. Then again, one can still write an explicit command, so that can be done at a later stage in a separate issue/PR.

siiptuo:
> How does this deal with the mis-usage when the variables exist but are empty?

This will ignore git/hg config and the user must specify `--copyright`. I think that this unset/override behavior is consistent with  other applications.

> Personally I would like a small log-statement that the environment-variables or git config is used, because this behavior is quite 'magical'. Perhaps something in the debug mode.

Good idea! This is now logged in debug mode.

> We would be careful with making this the default

> I don't see this work well in a more corporate environment where a company or organization is the copyright holder.

I agree that relying on git/hg doesn't work here. But I think that this is a reasonable default if it could be overridden with project-specific `.reuse/config` file as described in #13 and #68. This PR could be the first step in this direction.

nicorikken:
@siiptuo wauw, thanks for your work. For me this seems good to merge. I have yet to try it though, but the code looks correct and is well structured and tested.

mxmehl:
Thanks for the contribution! From what I can see, the code looks fine.

However, I'll have to test this properly first, and also think about possible implications and how to document this to avoid confusion and wrong header entries. I didn't have time to do that yet, so please apologise the delay.

mxmehl:
Sorry for returning to this so late. There have been a number of proposal for sources of information that REUSE shall include, often conflicting, and often causing a problem with various workflows.

Today, 5 of us sat down and tried to find a solution. Please see https://github.com/fsfe/reuse-tool/issues/68#issuecomment-1019229607 for the result.

Unfortunately, that means we won't take info from the Git/Mercurial config, although your code looks great! Instead, we are striving to support a project-wide share config file whose defaults a user can use or ignore.

PR #241, opened 2020-07-20
Validate file is readable in addheader subcommand

siiptuo:
Previously it was only checked that the given file is writable.  For example, non-existing files are writable so an exception was raised after validation when trying to open the file. To fix this, introduce "r+" mode to `PathType` which validates that the file is both readable and writable.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -44,6 +44,8 @@ Contributors
 
 - yoctocell
 
+- Tuomas Siipola
+
 
 Translators
 -----------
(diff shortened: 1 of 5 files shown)


nicorikken:
Nice small improvement!

mxmehl:
LGTM, thanks!

@carmenbianca would you like to have a look before merging?

carmenbianca:
Finally got to this. Awesome improvement—well spotted :)

PR #242, opened 2020-07-21
Use UTF-8 explicitly when reading files, too.

Flameeyes:
Since the file will be written out in UTF-8, it's a safe bet that they should also be read in UTF-8.

Alternatively, we could support a global setting to override the default encoding, but in 2020 that sounds like a recipe for more headaches than not.

### DIFF
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -384,7 +384,7 @@ def _add_header_to_file(
             out.write("\n")
             return result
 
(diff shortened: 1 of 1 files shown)


Flameeyes:
@carmenbianca ping? :)

mxmehl:
LGTM, thanks! @carmenbianca please revert if you see a serious issue with this.

PR #249, opened 2020-08-13
Detect comment style for Scheme (.scm) and ClojureScript (.cljs, .cljc) files

pukkamustard:
This adds entries to `EXTENSION_COMMENT_STYLE_MAP` for Scheme (.scm) and ClojureScript (.cljs, .cljc) files using `LispCommentStyle`.

### DIFF
--- a/src/reuse/_comment.py
+++ b/src/reuse/_comment.py
@@ -481,6 +481,7 @@ EXTENSION_COMMENT_STYLE_MAP = {
     ".rss": HtmlCommentStyle,
     ".sass": CssCommentStyle,
     ".scala": PythonCommentStyle,
+    ".scm": LispCommentStyle,
     ".scpt": AppleScriptCommentStyle,
     ".scptd": AppleScriptCommentStyle,
     ".scss": CssCommentStyle,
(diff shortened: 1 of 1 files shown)


pukkamustard:
I've been using this and I think it would make sense to instead use three semicolons for Scheme files (`;;;`).

This is more conventional (at least in Guile): https://www.gnu.org/software/guile/manual/html_node/Comments.html:

> use two semicolons for comments which are indented like code, and three semicolons for comments which start at column 0,

and what seems to be used in Guix (a large Guile project): https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/services.scm

This is just stylistic. But I think might be a strongish preference. Maybe this can be made configurable on a project-to-project basis? How about a template like system for comments?

mxmehl:
IIUC, you propose a new comment style for these files, right? If so, would you please open a new issue? This PR is rather simple, but your request is a different construction site.

pukkamustard:
Yes, I'm mixing up two things. Please ignore my second comment.

mxmehl:
Alright, please feel free to open a separate issue about the new comment style :)

PR #251, opened 2020-08-14
Add Meson's meson.build to the supported files.

Flameeyes:
The comment syntax is documented at https://mesonbuild.com/Syntax.html#comments and matches Python's.


mxmehl:
Thank you!

PR #254, opened 2020-09-10
add rst support

rscohn2:


mxmehl:
Thank you, looks good to me!

However, I saw that [single-line comments in RST are discouraged](https://stackoverflow.com/a/4783854/4273755). Do you think this is a problem in practice?

rscohn2:
It's not a problem. I checked the [rst spec](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#comments) and then checked each of the possible ambiguous cases. The SPDX header does not match any of them. It would be a problem if you used one of '|', '[', or '_' after the '..'. Or if you used '::' instead of ':'.

I tried to implement the multi-line, but was getting this error:
reuse addheader: error: 'test.rst' does not support multi-line comments, please do not use --multi-line

mxmehl:
I am not sure how to implement multiline comments in the tool, that's where @carmenbianca would have to help, but she's currently not available.

I'll merge this therefore, thanks for the contribution and explanation!

Issue #256, opened 2020-09-20
Add `--verbose` flag to lint, showing copyright/license info sources


robinkrahl:
If a project uses multiple licenses, it is hard to get an overview of which files are under which license without manually inspecting the files.  `reuse lint` only lists all used licenses, and while `reuse spdx` lists the license information for all files, it is rather hard to read for a human.

I think it would be good to have a `reuse list`command that prints a table with the copyright information for all files (or maybe `report` to increase the edit distance from `lint`) :

```
$ reuse list
File	License		Copyright Text
1.c	MIT		2020 John Doe
2.c	MIT		2020 John Doe
3.txt	CC0-1.0		2019 Jane Smith
```

It would be even better if the command could group files with the same copyright information:

```
$ reuse list
File	License		Copyright Text
*.c	MIT		2020 John Doe
3.txt	CC0-1.0		2019 Jane Smith
```

Use cases:
1. Getting an overview of a third-party project.
2. Making sure that you added the correct annotations for your own project.

I can try to implement this if you would merge it.

mxmehl:
Indeed, that'd be a nice view, also for debugging.

Perhaps related to #183 since `reuse spdx` should already include all information. Outputting it as JSON (and/or YAML) would allow for a simpler manipulation that could also be done manually, e.g. with tools like `jq`.

robinkrahl:
Yes, using a separate parser for the `reuse spdx` output would also be a way to implement this.  Does this mean you’d rather not have this feature in reuse-tool directly?

mxmehl:
> Does this mean you’d rather not have this feature in reuse-tool directly?

I am fine with having this function, either as a full list (as per your first post), or also with the option to have a detailed view for a single file (e.g. explaining the chain of precedence in case there are multiple explicit/glob coverages for this file).

Regarding spdx, I just wanted to make sure that we do not forget that users also may have this function to build their own custom query if necessary.

mxmehl:
In today' hackathon we concluded that we would like to have this verbose listing as part of the `lint` subcommand, e.g. `reuse lint --verbose`, perhaps also combinable with `--json`.

Aside from the detected license and copyright it should also provide info on *where* this information has been derived from, e.g. the file itself, the .license file, or dep5 (or future REUSE.yaml).

mxmehl:
So us maintainers would like to see this feature in the next release.

What's unclear yet:
[...]

Issue #183, opened 2020-02-19
Add `--json` flag to lint


carmenbianca:
All the infrastructure for this is already in place, but it could be really helpful for people who want to programmatically automate stuff: a `reuse json` interface. It'd simply output a combination of `reuse lint` and `reuse spdx`, but in JSON format.

carmenbianca:
I've had a longer think about this.

`reuse` has an internal representation for its data. This data is dictionary-like, and can very easily be transformed into JSON. However, if the internals were somehow changed, so would the output. This is *not* ideal as an outward API. This could be circumvented by implementing a bridge between the new internals and the formerly-internal API, but that's a bit of a stupid bridge:

- The internal structure probably isn't the best way to represent the data *anyway*.
- We would have to document the JSON API.
- People really dislike having to learn more APIs, and usually just divine the API from the output.
- To quote Raymond Hettinger: [There must be a better way!](https://blog.vnmis.net/there-must-be-a-better-way/).

So I recently learnt that SPDX soon wants to adopt JSON and YAML as new output formats. Or rather, they want to formalise those formats. @silverhook brought up that YAML is likely to become dominant among the output formats. I would personally prefer JSON over YAML because JSON is in the Python standard library, and is imo a much cleaner, clearer standard, but I can concede that YAML is a lot prettier to look at.

So instead of doing our own JSON API, it would be much nicer to simply output SPDX JSON/YAML. If the SPDX JSON/YAML doesn't contain some stuff that we need, we can:

1. add custom fields;
2. add a separate section/key at the bottom of the document;
3. add `reuse json` after all with a standalone API.

The way I see it, this could be implemented really well through <https://github.com/spdx/tools-python> as a library. We could contribute to that project to add some of the things we need, and remove some code in this repository in the process. As a plus, if we use that repository (and patch it up), we get multiple output formats for free. It'd be a matter of `reuse spdx --json` and `reuse spdx --yaml`.

The only problem, however, is that I am not entirely happy with `tools-python`. It's not documented very well, and it has some silly things going on. I'd tally them, but I'll withhold for now.

I'm fairly confident, however, that tighter integration with an improved `tools-python` would improve this tool a lot.

mxmehl:
Interesting thoughts! One way to bring this issue on SPDX' table could be to write an email to the spdx-tech mailing list (https://lists.spdx.org/g/spdx-tech) and perhaps have this as a topic in one of their calls.

silverhook:
[...]

PR #258, opened 2020-09-22
Add list command to list project files and licenses

robinkrahl:
This patch adds the list command that lists all project files and their
licenses in a human-readable format.

---

This PR provides a basic implementation of the list command as discussed in issue #256.  (It does not use the name report as there already is a `reuse.report` module.)

There are some question I’d like to discuss with you:
- Should we include additional information, for example the copyright text?  This might be hard to read in a table, especially if the text contains new lines.  We could replace new lines with spaces.
- Should we perform any kind of magic with the paths?  For example, trying to shorten them if they are longer than e. g. 80 characters, or grouping similar paths with the same license information.

Note that I had to deactive the pre-commit hook because it caused this error:

```
black....................................................................Passed
pylint...................................................................Failed
- hook id: pylint
- exit code: 1

  File "/home/robin/reps/reuse-tool/venv/bin/pylint", line 10, in <module>
    sys.exit(run_pylint())
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/pylint/__init__.py", line 19, in run_pylint
    from pylint.lint import Run as PylintRun
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/pylint/lint/__init__.py", line 75, in <module>
    from pylint.lint.check_parallel import check_parallel
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/pylint/lint/check_parallel.py", line 7, in <module>
    from pylint import reporters
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/pylint/reporters/__init__.py", line 24, in <module>
    from pylint import utils
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/pylint/utils/__init__.py", line 46, in <module>
    from pylint.utils.ast_walker import ASTWalker
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/pylint/utils/ast_walker.py", line 6, in <module>
    from astroid import nodes
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/astroid/__init__.py", line 47, in <module>
    from astroid import inference, raw_building
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/astroid/inference.py", line 34, in <module>
    from astroid import (
  File "/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/astroid/protocols.py", line 29, in <module>
    from astroid import Store, arguments, bases
ImportError: cannot import name 'Store' from 'astroid' (/home/robin/reps/reuse-tool/venv/lib/python3.7/site-packages/astroid/__init__.py)
```

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-"""Tests for reuse._main: lint, spdx, download"""
+"""Tests for reuse._main: lint, list, spdx, download"""
 
 # pylint: disable=redefined-outer-name,unused-argument
 
@@ -139,6 +139,53 @@ def test_lint_no_multiprocessing(fake_repository, stringio, multiprocessing):
 
     assert result == 0
     assert ":-)" in stringio.getvalue()
+
+
+def test_list(fake_repository, stringio):
+    """List all files and their copyright information."""
+    os.chdir(str(fake_repository))
+    result = main(["list"], out=stringio)
+
+    lines = [line.rstrip() for line in stringio.getvalue().split("\n")]
+    expected_lines = [
+        "File                    License",
+        "----------------------  ----------------------------------------",
+        "doc/index.rst           CC0-1.0",
+        "src/custom.py           LicenseRef-custom",
+        "src/exception.py        Autoconf-exception-3.0, GPL-3.0-or-later",
+        "src/source_code.c       GPL-3.0-or-later",
+        "src/source_code.html    GPL-3.0-or-later",
+        "src/source_code.jinja2  GPL-3.0-or-later",
+        "src/source_code.py      GPL-3.0-or-later",
+        "",
+    ]
+
+    assert result == 0
+    assert lines == expected_lines
+
+
+def test_list_no_multiprocessing(fake_repository, stringio):
+    """List all files and their copyright information, with the
+    --no-multiprocessing option set."""
+    os.chdir(str(fake_repository))
+    result = main(["--no-multiprocessing", "list"], out=stringio)
+
+    lines = [line.rstrip() for line in stringio.getvalue().split("\n")]
+    expected_lines = [
+        "File                    License",
+        "----------------------  ----------------------------------------",
+        "doc/index.rst           CC0-1.0",
+        "src/custom.py           LicenseRef-custom",
+        "src/exception.py        Autoconf-exception-3.0, GPL-3.0-or-later",
+        "src/source_code.c       GPL-3.0-or-later",
+        "src/source_code.html    GPL-3.0-or-later",
+        "src/source_code.jinja2  GPL-3.0-or-later",
+        "src/source_code.py      GPL-3.0-or-later",
+        "",
+    ]
+
+    assert result == 0
+    assert lines == expected_lines
 
 
 def test_spdx(fake_repository, stringio):
--- a/src/reuse/_main.py
+++ b/src/reuse/_main.py
@@ -18,6 +18,7 @@ from . import (
     header,
     init,
     lint,
+    list,
     spdx,
 )
 from ._format import INDENT, fill_all, fill_paragraph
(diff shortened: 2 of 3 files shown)


robinkrahl:
I’m still interested in this feature.  Is there anything I can do to move on with this PR?

mxmehl:
Thanks for your patch! I finally found the opportunity to test it locally.

>     * Should we include additional information, for example the copyright text?  This might be hard to read in a table, especially if the text contains new lines.  We could replace new lines with spaces.

Yes, I think we need this. A table may be a quite unsuitable format anyway for file paths and such often lenghty information, but probably the most user-friendly for a quick grep.

>     * Should we perform any kind of magic with the paths?  For example, trying to shorten them if they are longer than e. g. 80 characters, or grouping similar paths with the same license information.

I see a bonus regarding fanciness, but if the main goal of the table format is to ease `grep`-ing things, this would not really help I am afraid.

So in general I think the practical usefulness would be higher if we (also) output the information as JSON or YAML, and also include the full copyright line(s) with it.

Another point: Right now, if license information is missing for a file, the column is just empty. Shall we write something like `<missing>` or so instead?

> Note that I had to deactive the pre-commit hook because it caused this error:

Strange. Let's see how we can fix that later.

robinkrahl:
Thanks for the feedback!

Actually, my intention with this command is to provide a human-readable
overview over the license situation, but I see that it might also be
interesting for automated analysis.

So my suggestions would be:
1. Keep the table as the default output format intended to be read by
   humans.
2. Add an option for setting the output format to TSV (for grepping,
   cutting etc.) and JSON (for scripting).
3. Add an option for selecting the information to display (i. e. whether
   to include the copyright text or not).

Does that sound reasonable?

mxmehl:
> Does that sound reasonable?

Yes, absolutely. I think we have to think all these options as one package or subcommand so that we do not bar ourselves in subsequent changes.

My current idea for possible command line options:

```
# output licensing and copyright for all files in different formats
reuse list --tsv      # table
reuse list --csv      # comma-separated (is this necessary?)
reuse list --json     # json
reuse list --yaml     # yml

reuse list --tsv --hide-copyright   # as table, but no copyright column
reuse list --json --hide-license    # as json, but no license column
```

However, I also think we have to keep #183 and #320 in mind. A `--json` flag is certainly useful for `lint` and `spdx` as well. I am not sure whether it makes sense to combine these output options programmatically, but we should decide this before working on `list` separately.

weeman1337:
Hey. What about changing the options to something like this?

```
--format=json
--format=csv --separator=";"
```

With `--format` it isn't possible to define two at a time like `--json --tsv` would allow. This reduces the logic in the command handler. Furthermore in the `csv` case you would be free to choose the separator that you need.

tests/test_main.py
```diff
@@ -141,6 +141,53 @@ def test_lint_no_multiprocessing(fake_repository, stringio, multiprocessing):
...
 
 
+def test_list(fake_repository, stringio):
+    """List all files and their copyright information."""
+    os.chdir(str(fake_repository))
+    result = main(["list"], out=stringio)
+
+    lines = [line.rstrip() for line in stringio.getvalue().split("\n")]
+    expected_lines = [
+        "File                    License",
```
weeman1337, line 151: Maybe there should be an option to skip the headers.
Makes things easier e.g. if one processes the results by a script.

mxmehl:
@weeman1337's suggestions make sense. I'd love to see this feature merged, but my questions in the [earlier comment](https://github.com/fsfe/reuse-tool/pull/258#issuecomment-828284965) are still open and unclear to me.

mxmehl:
Please see https://github.com/fsfe/reuse-tool/issues/256#issuecomment-1019240709

It would rather make sense to put this functionality in `lint` rather than a separate subcommand, and also providing additional information.

Therefore, although this PR could be very helpful to implement this slightly different goal, I'd close this PR in order to keep things clean. Thanks so far already!

PR #259, opened 2020-09-22
default_template: follow Kernel style

aparcar:
The Kernel suggests to put the SPDX license identifier as high up as
possible, reflect that in the default_template.

[...]
The SPDX license identifier in kernel files shall be added at the first
possible line in a file which can contain a comment. For the majority or
files this is the first line, except for scripts which require the
‘#!PATH_TO_INTERPRETER’ in the first line. For those scripts the SPDX
identifier goes into the second line.
[...]

Signed-off-by: Paul Spooren <mail@aparcar.org>

### DIFF
--- a/src/reuse/templates/default_template.jinja2
+++ b/src/reuse/templates/default_template.jinja2
@@ -1,7 +1,8 @@
+{% for expression in spdx_expressions %}
+SPDX-License-Identifier: {{ expression }}
+{% endfor %}
+
 {% for copyright_line in copyright_lines %}
 {{ copyright_line }}
 {% endfor %}
 
-{% for expression in spdx_expressions %}
-SPDX-License-Identifier: {{ expression }}
-{% endfor %}


mxmehl:
Thank you for your contribution.

However, I wonder whether it is a good idea to change the default template because of one project's preferences (OK, an important one, but still).

Would it be sufficient for your work to change the template on your side? [Here's](https://reuse.readthedocs.io/en/stable/usage.html#templates) some help on how to do that.

aparcar:
Are there other project having the current style as a preference? I'd guess that SPDX license tags are more important than copyright owners, especially for parsing scripts, so having it as early as possible seems reasonable.

Also the Kernel is likely a more influential project regarding style and consistency than usual project.

Anyway, I can also deal with a custom template, it's up to you.

In case this is of interest I'll fix the failing tests.

mxmehl:
Since there are a lot of projects using REUSE already, I'd like to avoid changing the current behaviour. Also, we'd have to change docs and the tutorial to match the expected output with the actual.

Kernel is using their own scripts AFAIK, and in many other ways it's an exceptional project, so I'm fine with that :)

I hope that templates suit your needs then.

PR #267, opened 2020-10-22
Add screencast for tool commands

mxmehl:
This adds a ~1min GIF with a demo of a few commands to make the example repository REUSE compliant.

I've created this with [demo-magic](https://github.com/paxtonhare/demo-magic) and SimpleScreenRecorder, so we can easily change commands and create screencasts for other procedures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ inspection of copyright and licenses in software projects. The REUSE
 helper tool, on the other hand, is solely designed to be a simple tool
 to assist in compliance with the REUSE recommendations.
 
+## Example demo
+
+In this screencast, we are going to follow the
+[tutorial](https://reuse.software/tutorial), making the [REUSE example
+repository](https://github.com/fsfe/reuse-example/) compliant.
+
+![Demo of some basic REUSE tool commands](https://download.fsfe.org/videos/reuse/screencasts/reuse-tool.gif)
+
 ## Install
 
 ### Installation via pip


ferdnyc:
@mxmehl You don't happen to still have the demo-magic input file for this, do you?

I'd like to re-capture the screencast, because currently the image size is large enough that both GitHub and ReadTheDocs are scaling it down where it's embedded in the documentation. Turns out that makes the font a _lot_ harder to read. (Particularly the green text, which grows considerably dimmer.)

I think generating a new version using smaller terminal dimensions and/or a larger font size will improve the readability greatly.

I can of course just re-create the scripting, but if you've still got the original version then I get to be lazier, **and** we get a new GIF that's identical to the current screencast in terms of content/timing/etc.

ferdnyc:
(Heh. I got my hopes up when I found https://wiki.fsfe.org/Teams/REUSE/VideoScript , but turns out that's the spoken-word script for an **actual** video production.)

mxmehl:
Sorry for the late reply. I found the script I've used, using [demo-magic](https://github.com/paxtonhare/demo-magic). This is the script I've (probably) used (but I'm not sure whether I may have slightly modified demo-magic):

```sh
#!/bin/bash

# You can choose different typed, e.g. pe or pei
TYPE=pe
# What to do with comments? : for doing nothing, $TYPE for doing the same as with code
COMM=":"

# Set path to demo-magic here
. ~/Repositories/github/demo-magic/demo-magic.sh

# Set path to the example repo
cd ~/Repositories/fsfe/reuse-example

git checkout noncompliant

clear

wait
$COMM "# Let's see the status of the repo"
$TYPE 'tree -a -I ".git"'
$TYPE 'reuse lint'
$COMM "# Now we are going to add information to all files in three steps:
# 1. GPL-3.0-or-later for main.c, Makefile, and README.md with copyright Jane Doe
# 2. CC-BY-SA-4.0 for all images, copyright Great Artist
# 3. CC-1.0 for the simple .gitignore file"
wait
$TYPE 'reuse addheader --copyright="Jane Doe <jane@example.com>" --license="GPL-3.0-or-later" src/main.c Makefile README.md'
#$TYPE 'git diff'
$TYPE 'reuse addheader --copyright="Great Artist <artist@example.net>" --license="CC-BY-4.0" img/*'
#$TYPE 'git status'
#$TYPE 'cat img/cat.jpg.license'
$TYPE 'reuse addheader --copyright="Jane Doe <jane@example.com>" --license="CC0-1.0" .gitignore'
$COMM "# So, are we good now?"
wait
$TYPE 'reuse lint'
$COMM "# Exactly, the license text files are missing! Let's download them..."
wait
$TYPE 'reuse download --all'
$TYPE 'reuse lint'
$COMM "# All good! Now, what about a SPDX bill of material?"
wait
$TYPE 'reuse spdx'

wait

git reset --hard
git clean -dfx
```

The original webm file is here, definitely has a better resolution: https://download.fsfe.org/videos/reuse/screencasts/

ferdnyc:
@mxmehl Great, thanks! I'll take a stab at regenerating the screencast at GitHub-approved resolution/dimensions.

Issue #271, opened 2020-11-18
Feature: Support PlantUML comments


wtraylor:
[PlantUML](https://plantuml.com) is a language for describing UML diagrams in text files. Typical file extensions are `.plantuml`, `.iuml`, `.puml`, and `.uml`. The [documentation](https://plantuml.com/de/commons#8413c683b4b27cc3) explains a rather special commenting syntax:

> Everything that starts with simple quote `'` is a comment.
>
> You can also put comments on several lines using `/'` to start and `'/` to end.

This comment style is not yet supported by the reuse-tool. It would be great to have it included.

rscohn2:
Here is a sample PR for adding support for fortran and bat files: https://github.com/fsfe/reuse-tool/pull/255/files

If you make the changes, you can test it with applications that read the files.

wtraylor:
Thanks. I will create a pull request.

PR #273, opened 2020-11-19
Add PlantUML comment style

wtraylor:
The extensions .plantuml, .puml, .iuml, and .pu are all associated with
PlantUML, according to fileinfo.com, file-extension.org, and
fileinfo.co.

The simple .uml extension is used by several other UML applications and
thus shouldn’t be hard-coded to PlantUML.

Fixes GitHub issue gh-271


mxmehl:
LGTM. Thank you!

PR #276, opened 2020-11-25
Use JsxCommentStyle for TSX files

mikkopiu:
.tsx is just a naming convention for TypeScript JSX files.


mxmehl:
Thank you for your contribution!

kirelagin:
Linters don’t seem to be particularly happy about this...

mikkopiu:
> Linters don’t seem to be particularly happy about this...

Yeah, after the fact I think using `CCommentStyle` would've probably been a better choice. I at least haven't come across any cases where using JSX syntax (`{/* */}`) in comments would be necessary (or preferable) over C-style comments (`//`). Maybe if the full file is just a JSX template instead of JSX templating within a JS/TS file?

PR #282, opened 2020-12-10
Use CCommentStyle for TS files

mikkopiu:
TypeScript (.ts) files use the same comment style as Javascript files (.js)


mxmehl:
Thank you for your contribution!

PR #284, opened 2020-12-13
Make addheader detect .dockerignore, Gemfile and Rakefile

paddy-hack:
I just made a Ruby on Rails project REUSE compliant and noticed the `Gemfile` and `Rakefile` were not covered yet.

I also like to keep my Docker container build contexts small so added `.dockerignore` as well.


mxmehl:
Great, thanks!

PR #294, opened 2021-01-13
feat: Support Jenkinsfile for comment style

nicorikken:
Support the syntax for the default Jenkins pipeline filename 'Jenkinsfile'. The
syntax is Groovy, so CCommentStyle is used (`//` and `/* ... */`). Have tested
this and it works.


mxmehl:
Thank you!

PR #295, opened 2021-01-13
Add support for .mjs files

ethulhu:
To distinguish JavaScript modules from non-module files, some sources recommend using the `.mjs` extension [1]. They are otherwise identical to JavaScript `.js` files.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_—_.mjs_versus_.js


nicorikken:
Nice one, looks good to me!

Issue #189, opened 2020-02-24
addheader should recognise file extensions that *require* .license files


carmenbianca:
e.g., .json.

nicorikken:
What would be a neat way to achieve this? Can there be a similar map like `EXTENSION_COMMENT_STYLE_MAP` in `src/reuse_comment.py` to check against? I notice that currently `.sqlite` and `.whl` files are already handled in a different way, but I'm not able to find out how that works.

mxmehl:
Regarding your last question, this happens via `is_binary` from the `binaryornot.check` module.

I am not sure whether we would come around adding file extensions manually. I am not aware of a list for such files.

nicorikken:
Alright, looking at the code, it makes sense to have an additional comment style which equates to `NoCommentStyle` possible. This can be the result of `_get_comment_style(path)`, as is being called in `_add_header_to_file()` in `header.py`. So following the execution flow of `run()` in `header.py` first the binary files are handled differently, and then plaintext files are handled. And in the process of plaintext files the `NoCommentStyle` files are then dealt with. Similar to the shortcut in `run()` in `header.py` the `path` variable can be replaced to represent the `<filename>.license` path, and the execution can then be followed. In this case the style also has to be changed to an `EmptyCommentStyle` so the correct details are added in the new `.license` file.

I hope my worded PR is clear enough :smile:  Either way I now have an idea how to implement this.

mxmehl:
A better Python programmer than I am would probably understand this better, but I think I can follow your argumentation well enough, thanks :)

mxmehl:
Closed with #298, thanks @oxzi!

PR #298, opened 2021-01-14
Comment style for uncommentable files

oxzi:
Previously, only binary files were handled as uncommentable [0]. For
other files the "--explicit-license" flag has to be set explicitly.
However, there are also text files which are uncommentable.

This commit / pull request introduces support for a new pseudo comment
style for uncommentable text files.

One use case is, for example, Golang's go.sum file. Hashes of the
referenced modules are listed there, but without any special header [1].
Thus it is not possible to add an SPDX header. However, this file cannot
be omitted. Whether the file is to be ignored license-wise was addressed
in a previous pull request [2], but not clarified.

With the changes introduced here, this file is recognized as
uncommentable and a .license file is created.

[0] https://reuse.software/tutorial/#binary-and-uncommentable-files
[1] https://blog.golang.org/using-go-modules
[2] https://github.com/fsfe/reuse-tool/pull/234

### DIFF
--- a/src/reuse/_comment.py
+++ b/src/reuse/_comment.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: 2019 Kirill Elagin
 # SPDX-FileCopyrightText: 2020 Dmitry Bogatov
 # SPDX-FileCopyrightText: 2021 Alliander N.V.
+# SPDX-FileCopyrightText: 2021 Alvar Penning
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -427,6 +428,12 @@ class TexCommentStyle(CommentStyle):
     INDENT_AFTER_SINGLE = " "
 
 
+class UncommentableCommentStyle(EmptyCommentStyle):
+    """A pseudo comment style to indicate that this file is uncommentable. This
+    results in an external .license file as for binaries or --explicit-license.
+    """
+
+
 #: A map of (common) file extensions against comment types.
 EXTENSION_COMMENT_STYLE_MAP = {
     ".adb": HaskellCommentStyle,
@@ -488,6 +495,7 @@ EXTENSION_COMMENT_STYLE_MAP = {
     ".jinja": JinjaCommentStyle,
     ".jinja2": JinjaCommentStyle,
     ".js": CCommentStyle,
+    ".json": UncommentableCommentStyle,
     ".jsx": JsxCommentStyle,
     ".jy": PythonCommentStyle,
     ".ksh": PythonCommentStyle,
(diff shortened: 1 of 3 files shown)


nicorikken:
@oxzi nice one! You beat me to it, as I already did the thinking for it this morning in #189 I like your approach better, with naming it 'uncommentable' and checking all paths beforehand. I was weary that the check might take to long, then again it is only a short list.

So, this looks very good to me and a worthwhile improvement.

It would be nice if we could have a test to show that `go.sum` is considered Uncommentable and thus a new file is created. Not sure how that fits in with the current tests, I haven't looked into that yet.

mxmehl:
That's an interesting approach indeed, thank you!

I've just tested it by adding a `go.sum` file, and this is the output. Shouldn't `go.sum` be detected automatically?

```
reuse addheader --license "MIT" --copyright "Me" go.sum
usage: reuse addheader [-h] [--copyright COPYRIGHT] [--license LICENSE] [--year YEAR]
                       [--style {applescript,aspx,bat,bibtex,c,css,f,haskell,html,jinja,jsx,lisp,m4,ml,plantuml,python,rst,tex}]
                       [--copyright-style {spdx,string,string-c,string-symbol,symbol}]
                       [--template TEMPLATE] [--exclude-year] [--single-line] [--multi-line]
                       [--explicit-license] [--skip-unrecognised]
                       path [path ...]
reuse addheader: error: 'go.sum' does not have a recognised file extension, please use --style, --explicit-license or --skip-unrecognised
```

For testing purposes, I also added `.json` to `EXTENSION_COMMENT_STYLE_MAP`. Unfortunately, the output is the same.

oxzi:
@nicorikken: Thanks for your feedback. I tried to look for related issues/PRs but it seems like I used the wrong keywords. And of course I can go back and look at the tests.

@mxmehl: I tried reproducing my changes and it worked for me both with my local Python environment as well as with a generated Docker container based on the `Dockerfile`. Is it possible that you are using a wrong `reuse` executable?

```
$ git fetch upstream pull/298/head:pr-298
From https://github.com/fsfe/reuse-tool
 * [new ref]         refs/pull/298/head -> pr-298

$ git checkout pr-298
Switched to branch 'pr-298'

$ git rev-parse HEAD
339aa9c09ab88f3deb13e59a9d4ab7115b441b15

$ python setup.py install
...

$ touch go.sum

$ reuse addheader --license MIT --copyright Me go.sum
Successfully changed header of go.sum.license

$ ls go.sum*
go.sum  go.sum.license
```

oxzi:
I have updated the PR to also mark JSON files as uncommentable. Furthermore, a test based on a JSON file is now also included.

nicorikken:
Nice! I have tested your branch locally and I see that the json files now receive their own `.license` file. This is great! Looks good to merge for me.

mxmehl:
> @mxmehl: I tried reproducing my changes and it worked for me both with my local Python environment as well as with a generated Docker container based on the Dockerfile. Is it possible that you are using a wrong reuse executable?

Sorry for the false alarm. Somehow my local dev environment failed me. Can now reproduce that everything works well.

Also thanks for writing a test, excellent!

oxzi:
Thanks to both of you for the quick testing and merging.

PR #301, opened 2021-01-15
Ignore .hgtags file

nicorikken:
In order to solve issue https://github.com/fsfe/reuse-tool/issues/227

.hgtags is not a file that is managed by humans, so it makes sense to ignore it.


mxmehl:
Thanks!

Issue #306, opened 2021-01-25
MANIFEST.in support


nicorikken:
I encountered a project where a `MANIFEST.in` file was located. This is a file that is typically present for compiling Python source distributions: https://packaging.python.org/guides/using-manifest-in/

The current `_comment.py` already contains support for `Manifest.in`. It seems this was intended to be `MANIFEST.in`, but I'm not sure. Here is the actual line: https://github.com/fsfe/reuse-tool/blob/af5467130298e483db28624c05edbe5ebaeeae48/src/reuse/_comment.py#L593
Should I add a duplicate entry for `MANIFEST.in` in capital letters, or should I replace the existing entry?

nicorikken:
A search through GitHub code only found me one example of `Manifest.in` in hunderds of hits. Most importantly the file was empty and not being used.
Also I did a test locally and found that `MANIFEST.in` is picked up automatically, and `Manifest.in` is ignored by `python setup.py sdist`.

So I conclude that it should actually be `MANIFEST.in` and I will create a pull request to match that change.

Perhaps original authors of that change @carmenbianca or @KAction can correct me if I'm wrong.

carmenbianca:
Closed by #307 

Thanks!

PR #307, opened 2021-01-26
fix: comment support from Manifest.in to MANIFEST.in

nicorikken:
Correct the supported filename for Python source package configurations from
Manifest.in to the expected MANIFEST.in. Manifest.in is the default location for
this location, as is described in the Python packaging documentation:
https://packaging.python.org/guides/using-manifest-in/

This commit was preceded by a small consideration as documented in issue
https://github.com/fsfe/reuse-tool/issues/306


carmenbianca:
LGTM! Thanks for spotting this. I'll add a changelog entry.

PR #309, opened 2021-01-27
Add comment styles for Racket and Org files

yoctocell:
Use  `LispCommentStyle` for Racket, and `PythonCommentStyle` for Org files.


carmenbianca:
LGTM. Thanks a heap :)

PR #310, opened 2021-01-27
Bump requirements.txt and requirements-dev.txt

carmenbianca:
This bump keeps everything in sync with upstream dependencies. Pylint is being annoying as usual.


carmenbianca:
Apparently the latest version of python-debian has a Windows incompatibility. I'll report this upstream and declare a dependency on an earlier version in the meantime.

PR #317, opened 2021-02-02
Declare support for Python 3.10

carmenbianca:
Fairly simple :)

### DIFF
--- a/.github/workflows/pythonpackage.yaml
+++ b/.github/workflows/pythonpackage.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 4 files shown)


nicorikken:
It seems that the CI tooling is not equipped for 3.10 at the moment. Would that hurt other pull-requests? Other than that this seems good to me.
Is there an option to have a ">=3.6" option, so we don't have to create a similar pull-request each Python release? Then again it wouldn't hurt to explicitly verify that the new version is actually supported.

mxmehl:
Yes, I also tried to re-run the check but to no avail. So we'll have to wait as it seems.

>  Then again it wouldn't hurt to explicitly verify that the new version is actually supported.

Yes, that's also my gut feeling. Better check manually and make it available a bit later than causing avoidable issues.

PR #319, opened 2021-02-04
Add isort check to blackcheck

carmenbianca:
Previously `make blackcheck` could succeed even if the imports were not correctly ordered


mxmehl:
Do you see any problems with merging this @carmenbianca? Just wondering why you didn't do so already ;)

nicorikken:
Good addition!

Issue #238, opened 2020-06-22
Install git in Docker image?


m90:
Many CI providers allow the use of arbitrary Docker images for running certain tasks, but all of them require git to be installed in container so that they can do SCM related tasks like checking out tags et al.

Would it be an option to include git in the final `fsfe/reuse` image so that it can be used in CI directly?

---

NB: I'd be happy to add a PR for this, but wanted to ask beforehand if you even want it.

mxmehl:
Thanks for the suggestion. May I ask for a use case?

The reason why I am a bit hesitant is that the REUSE Docker image should be as slim as possible. It's being used in many CI pipelines, and to reduce build time and/or disk/download space, a large image should be avoided.

m90:
So my exact use case is using this image in a CircleCI pipeline where it fails when you build a Git Tag which gives me:

```
Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
Enumerating objects: 1014, done.
Counting objects: 100% (1014/1014), done.
Compressing objects: 100% (427/427), done.
Total 12542 (delta 599), reused 936 (delta 550), pack-reused 11528

object not found
```

If I use an image that has git installed and use pip to install `reuse` it works fine. I did not try `apk add git` in the `fsfe/reuse` container though yet. Maybe that would be the safest way for me.

carmenbianca:
https://github.com/fsfe/reuse-tool/blob/master/Dockerfile

Git is installed. It's probably SSH that's missing.

> So my exact use case is using this image in a CircleCI pipeline where it fails when you build a Git Tag

Can you link to the circleci configuration file for this?

m90:
Isn't git just installed in the builder stage and will be excluded from the final image as it's not copied over?

This is the circle config that failed on the tag (branches would work): https://github.com/offen/offen/blob/14244a5eda0fb2cb6fbc5129abc4912be18306ab/.circleci/config.yml

This is our current fix that works against branches and tags: https://github.com/offen/offen/commit/f3cad81e5ea88b8296e1c909f2fabf874002971e#diff-1d37e48f9ceff6d8030570cd36286a61

carmenbianca:
Found it. Normally in a CI system, you take the following steps:

1. Clone (and prepare) the repo.
2. Activate the Docker image.
3. Do the thing that is specific to the Docker image.

[...]

PR #321, opened 2021-02-07
Improve extra Docker image, add a Debian-based

mxmehl:
This makes the tag `latest-extra` build like the normal file with the extra packages, and not use `fsfe/reuse:latest` as base.

Additionally, it creates a new Dockerfile using `python:slim-buster` as base. It's ~200% larger but should be easier to comply with as alpine has its issues in this regard.

### DIFF
--- a/Dockerfile-debian
+++ b/Dockerfile-debian
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2021 Free Software Foundation Europe e.V. <https://fsfe.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Like normal Dockerfile, but based on python:slim (Debian) to ease compliance
+
+# Create a base image that has dependencies installed.
+FROM python:slim AS base
+
+RUN apt-get update \
+    && apt-get install -y git mercurial \
+    && rm -rf /var/lib/apt/lists/*
+
+# Build reuse into a virtualenv
+FROM base AS build
+
+WORKDIR /reuse-tool
+
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+COPY . /reuse-tool/
+
+RUN pip3 install -r requirements.txt
+RUN pip3 install .
+
+
+# Copy over the virtualenv and use it
+FROM base
+COPY --from=build /opt/venv /opt/venv
+
+ENV VIRTUAL_ENV=/opt/venv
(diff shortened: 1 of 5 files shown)


mxmehl:
Ping @zvr :)

Dockerfile-debian
```diff
@@ -0,0 +1,39 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Like normal Dockerfile, but based on python:slim (Debian) to ease compliance
+
+# Create a base image that has dependencies installed.
+FROM python:slim-buster AS base
+
+RUN apt-get update \
+    && apt-get install -y git mercurial python3 \
```
zvr, line 11: I'm pretty confident that when you start from `python` base image you do *not* need to install python ;-)
This will save you some disk space (due to all the dependencies you are currently installing)

Dockerfile-debian
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: 2021 Free Software Foundation Europe e.V. <https://fsfe.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Like normal Dockerfile, but based on python:slim (Debian) to ease compliance
+
+# Create a base image that has dependencies installed.
+FROM python:slim-buster AS base
```
zvr, line 8: Why not simply `python:slim` ?

Dockerfile-extra
```diff
@@ -2,7 +2,36 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-# Use fsfe/reuse:latest as a base, and install additional packages on top
-FROM fsfe/reuse:latest
+# Like normal Dockerfile, but install additional packages on top
 
-RUN apk --no-cache add openssh-client
+# Create a base image that has dependencies installed.
+FROM alpine:3.11 AS base
```
zvr, line 8: Ah, is this not switched to `python` base, as well?

zvr (changes requested):
Not sure I understood the use of the -extra Dockerfile, but it's not important.

Dockerfile-debian
mxmehl: Oh sure, didn't see this tag

Dockerfile-debian
mxmehl: Whoops, my bad. Indeed, it saves ~30MB disk space

Dockerfile-extra
mxmehl: The idea here is that this is like the normal Dockerfile (based on alpine) just with a few extra packages installed. See #238 for background.

PR #326, opened 2021-03-03
Avoid failing when --all and --output are both given

maurooaranda:
Since --output can't be used with more than one license, it makes sense to
give a similar treatment to the combination of --all and --output, since --all is
likely to attempt downloading more than one license.

My suggestion is to ignore --output if --all is given, and warn the user.

Best regards,
Mauro.

### DIFF
--- a/src/reuse/download.py
+++ b/src/reuse/download.py
@@ -6,6 +6,7 @@
 
 import errno
 import sys
+import logging
 from gettext import gettext as _
 from os import PathLike
 from pathlib import Path
@@ -21,6 +22,8 @@ from ._util import (
 )
 from .project import Project
 from .report import ProjectReport
+
+_LOGGER = logging.getLogger(__name__)
 
 # All raw text files are available as files underneath this path.
 _SPDX_REPOSITORY_BASE_URL = (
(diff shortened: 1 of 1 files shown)


src/reuse/download.py
```diff
@@ -124,6 +127,11 @@ def _successfully_downloaded(destination: PathLike):
         # TODO: This is fairly inefficient, but gets the job done.
         report = ProjectReport.generate(project)
         licenses = report.missing_licenses
+        if args.file:
+            _LOGGER.warning(
+                _("--output has no efect when used together with --all")
```
mxmehl, line 132: Typo: efect -> effect

mxmehl (changes requested):
Thanks for this, good catch!

One typo found, otherwise fine with me.

src/reuse/download.py
maurooaranda: > One typo found, otherwise fine with me.

Oops, sorry about that.  Typo fixed.

mxmehl: Thanks! :)

mxmehl (approved):

PR #327, opened 2021-03-05
comment: add comment style for Bitbake files

Liambeguin:
Bitbake recipes and classes (.bb, .bbappend, and .bbclass) use a
combination of Python and Shell.
Bitbake configuration files (.conf) use the standard conf file syntax.

Link: https://www.yoctoproject.org/docs/latest/bitbake-user-manual/bitbake-user-manual.html#Concepts

Signed-off-by: Liam Beguin <liambeguin@gmail.com>


src/reuse/_comment.py
```diff
@@ -450,13 +450,17 @@ class UncommentableCommentStyle(EmptyCommentStyle):
...
+    ".bbappend": PythonCommentStyle,
+    ".bbclass": PythonCommentStyle,
     ".bib": BibTexCommentStyle,
     ".c": CCommentStyle,
     ".cl": LispCommentStyle,
     ".clj": LispCommentStyle,
     ".cljc": LispCommentStyle,
     ".cljs": LispCommentStyle,
     ".cmake": PythonCommentStyle,  # TODO: Bracket comments not supported.
+    ".conf": PythonCommentStyle,
```
mxmehl, line 463: I am not sure whether we can assume that all `.conf` files can be commented in the Python style. Is there some standard for this?

Liambeguin: I searched more and couldn't find an actual standard, only that it's widely used with this style.

Any ideas where I could search for such a standard?

mxmehl: Yeah, and even if there was one, it seems there are also cases where this would break. My suggestion would be to not define this, same for other files like `cfg` or so.

Liambeguin: Understood, I'll resend without the `.conf` change.

mxmehl: Excellent, thanks!

PR #331, opened 2021-03-12
Fix format call in sdpx_info_of

maurooaranda:
Hi,

I noticed a little error in one of the calls to format, in `spdx_info_of`.

With the current code, if that call gets executed it will result in a `KeyError` exception.


nicorikken:
Great find! Looks good to me.  I don't see a reason for a test.

mxmehl:
Thank you, can confirm that it avoid the KeyError with erroneous SPDX expressions!

PR #332, opened 2021-03-14
Add support for Monorepos

ypid:
Imaging a [monorepo] like this:

```
monorepo-example
├── project_a
│   ├── README.md
│   └── .reuse
│       └── dep5
├── project_b
│   ├── README.md
│   └── .reuse
│       └── dep5
└── README.md
```

Without this change, `reuse` is confused that at the root of the project, there is no `.reuse` directory structure.

[monorepo]: https://en.wikipedia.org/wiki/Monorepo

This PR probably needs discussion before it can be merged. I just went ahead and provided this POC implementation so that we have some code to talk about. At the very least, a test will need to be added. Refer to https://github.com/ypid/latex-packages for a real life (mono)repo for testing.

### DIFF
--- a/src/reuse/vcs.py
+++ b/src/reuse/vcs.py
@@ -196,18 +196,38 @@ class VCSStrategyHg(VCSStrategy):
         return None
 
 
+def find_dot_reuse_dir(cwd: PathLike = None, break_path: PathLike = None) -> Optional[Path]:
+    if cwd is None:
+        cwd = Path.cwd()
+    if break_path is not None:
+        break_path = Path(break_path).resolve()
+
+    for path_to_check in [cwd] + list(cwd.parents):
+        if break_path is not None and break_path.samefile(Path(path_to_check).resolve()):
+            break
+
+        if (path_to_check / '.reuse').is_dir():
+            return Path(os.path.relpath(path_to_check))
+
+    return None
+
+
 def find_root(cwd: PathLike = None) -> Optional[Path]:
     """Try to find the root of the project from *cwd*. If none is found,
     return None.
 
     :raises NotADirectoryError: if directory is not a directory.
     """
+    root = None
     if GIT_EXE:
         root = VCSStrategyGit.find_root(cwd=cwd)
-        if root:
-            return root
-    if HG_EXE:
+    if HG_EXE and root is not None:
         root = VCSStrategyHg.find_root(cwd=cwd)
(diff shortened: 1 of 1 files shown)


mxmehl:
Phew, thanks for bringing that up. However, I am a bit afraid of adding this as it would complicate the spec and also the tool. In our spec, we only allow for one .reuse directory: https://reuse.software/spec/#dep5

However, we're working on supporting more flexible files for bulk-licensing stuff that shall soft-deprecate the DEP5 files. See more details here: spdx/spdx-spec#502

If we introduced that, it should solve your issue, and we can dep5 deprecate slowly. What do you think?

ypid:
The proposal from https://lists.fsfe.org/pipermail/reuse/2020q3/000085.html sounds good. It would address it.

mxmehl:
Thanks for the info. I hope we'll have this implemented soon :)

adisbladis:
I struggled with adopting `reuse` for a [monorepo I am hacking on](https://github.com/nix-community/trustix) with quite a complex set of licenses and ended up writing a tool that invokes `reuse addheader ...` and generates a dep5 file.

Quoting myself from [my commit message](https://github.com/nix-community/trustix/pull/63/commits/78197520903d3336fd64cf19b1a90657088feaad):

> Writing reuse dependency files is not a fun excercise in a complex repository with multiple
sub packages with a complex set of different licenses and sometimes even different licenses
within the same directory.
> 
> For example we license Nix expressions under MIT no matter where in the tree they live
so that they can freely be copy-pasted into nixpkgs and other third party repos
while the most important bits are all GPL-3.0-only.
> 
> This script provides a way to express our rules and outputs a reuse compatible dep5 file

It's quite a different approach to licensing management, instead of checking that files have licensing attached it [declaratively manages the entire repository licensing](https://github.com/nix-community/trustix/pull/63/commits/78197520903d3336fd64cf19b1a90657088feaad#diff-00b3825cfc7ca0461b0ff20726befe34af5f98d816f3de569b5377e716e8f4b9R152-R342).

@mxmehl Would you be interested in integrating something like this into `reuse` itself? 
I would be happy to re-license my code if it's interesting to the reuse project.

ypid:
I just noticed that the use case I initially opened this PR for does not need code changes. Using `--root` for example `reuse --root . lint` in `project_a` works nicely.

PR #342, opened 2021-03-30
Explicitly require setuptools

hroncok:
`reuse/__init__.py` imports pkg_resources


mxmehl:
Sorry for the long wait, just merged it!

PR #328, opened 2021-03-08
Add option to merge copyright lines

Liambeguin:

PR #345, opened 2021-04-10
Make reuse init reuse info present in pyproject.toml

CharString:
I was using reuse on an existing project and was surprised `reuse init` didn't use the info from pyproject.toml, so I wrote a patch.

  See PEP 621 for the spec:
  https://www.python.org/dev/peps/pep-0621/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,7 @@ Jinja2==2.11.3
 license-expression==1.2
 python-debian==0.1.38
 requests==2.25.1
+toml==0.10.2
 
 setuptools==52.0.0
 setuptools-scm==5.0.1
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,8 @@ requirements = [
     "Jinja2",
     # Exactly what it says.
     "binaryornot",
+    # For reusing info from  pyproject.toml
+    "toml",
 ]
 
 test_requirements = ["pytest"]
--- a/src/reuse/init.py
+++ b/src/reuse/init.py
@@ -8,9 +8,10 @@ import sys
 from gettext import gettext as _
 from inspect import cleandoc
 from pathlib import Path
-from typing import List
+from typing import List, Mapping, IO, Union
 
 import requests
+import toml
 
 from ._licenses import ALL_NON_DEPRECATED_MAP
 from ._util import PathType, print_incorrect_spdx_identifier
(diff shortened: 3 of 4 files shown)


nicorikken:
Seems to be a good idea. I think this needs some coordination with #240 which focuses more on version control being the source for author details.
I think we need to consider in which file the information is gathered, to avoid having settings be loaded in both the `init.py` and `vcs.py` files.
Also I'm not sure that the project file can always be considered the source of truth. I have experienced it is created once and hardly updated. So maybe there need to be some option whether the source files or the project file get a priority. Some elegant merging like in #328 could help here.

CharString:
Maybe the invoker of  reuse should be considered as the source of truth and should reuse just collect all possible values from all sources and present them as possible defaults. Much like [pdm init](https://github.com/pdm-project/pdm) presents all python executables it can find, and asks which one to use and store in the config.

nicorikken:
@CharString I took a look at `pdm init`. So if I understand you correctly, you could be presented with a prompt like:

```
$ reuse addheader
No copyright holder provider, but found multiple options for copyright owner:
0. Jane Doe
1. J. Doe <j.doe@mailprovider.com>
2. MyCorp
Please select: 
```

Considering that the output of `pdm init` is saved, this could also be done for reuse, but that might be overkill for a simple selection.

CharString:
@nicorikken Yes, that's the idea. About saving the choices. I'm very new to `reuse`, and just started to use it on a toy project. The programmer in me will seek maximal automation and minimal repetitive action. But I'm not a lawyer and there are legal consequences to what reuse does, so my instinct says to err on presenting _explicit_ choices to the user of the tool. On the other hand, it's the vcs commit action that makes these choices final, so maybe we can be aggressive in automation and consider ourselves nothing more but a fancy editor macro.

nicorikken:
Great. About saving it, that would only make sens if you'd run it often. For example each git commit, or in a continuous integration process. I'm not sure if anybody is doing something like that. You could always provide explicit parameters to the `addheader` command to avoid having to make a selection each time.

So thinking about it know, I think an interactive selection would be nice if multiple options are encountered (like in this PR and #240 ). And saving it could be a future extension.

Regarding this specific PR, I'm not a REUSE maintainer, so it's not my call. I like the idea in general. This of course could grow to many types of project files, so we'd just have to make sure this is worth adding.

nicorikken:
The current line of thought is not to bloat REUSE with all sort of configuration options and suggest helper scripts instead: https://reuse.readthedocs.io/en/latest/scripts.html Perhaps instead of this pull request we can come up with a oneliner or small script to read author information from the pyproject.toml?

floriansnow:
@nicorikken Did you mean to close this PR with your last comment? If not, what would you say is the current state of this PR?

CharString:
@floriansnow I certainly considered it as closed by Nico :)

PR #347, opened 2021-04-15
Allow creating .license file for write-protected files

### LINKED ISSUE #346: Can’t write `.license` file for write-protected file
# Issue
`reuse` doesn’t create a separate `.license` file if the original file is write-protected/read-only.

# Steps to Reproduce

```bash
touch test.txt
chmod a-w test.txt
reuse addheader -l 'CC-BY-4.0' -c 'Jane Doe' --explicit-license test.txt
```

Error:

    reuse addheader: error: argument path: can't write to 'test.txt'

Environment:
- Reuse version: 0.12.1
- Operating system: ArchLinux
- Python version: 3.9.2

# Expected Behavior
`reuse` crates a `test.txt.license` file without error.

As long as the directory is writeable and a `.license` can be created, there should be no error.

CharString:
Fixes #346 

But relaxes a check on all path arguments passed to addheader command.
There are no tests for graceful permission handling yet.

So this is WIP. Just wanted to let people know someone picked up the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ The versions follow [semantic versioning](https://semver.org).
 
 - Fix faulty file types:
   - Extensible Stylesheet Language (`.xsl`) actually uses HTML comment syntax
+- Allow creating .license file for write-protected files (#347)
 
 ### Security
 
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -522,7 +522,7 @@ def add_arguments(parser) -> None:
         action="store_true",
         help=_("skip files with unrecognised comment styles"),
     )
-    parser.add_argument("path", action="store", nargs="+", type=PathType("r+"))
+    parser.add_argument("path", action="store", nargs="+", type=PathType("r"))
 
 
 def run(args, project: Project, out=sys.stdout) -> int:
--- a/tests/test_main_addheader.py
+++ b/tests/test_main_addheader.py
@@ -8,6 +8,7 @@
 
 # pylint: disable=unused-argument
 
+import stat
 from inspect import cleandoc
 
 import pytest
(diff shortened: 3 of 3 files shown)


mxmehl:
Thank you! We'll review your PR soon.

CharString:
I've been playing around with hypothesis, to generate tests that make sure the reuse cli handles all combinations of file permissions. But I bumped into a problem with its integration with pytest: HypothesisWorks/hypothesis#2939

mxmehl:
@CharString may I ask what's the status here? Is something in hypothesis still blocking the potential test and merge of this PR?

CharString:
Well. The test included in this PR tests whether the case from the issue works. But when coding this I noticed the code I changed, wasn't covered by tests. So we have no way to tell whether this breaks anything. That's why I wanted to use hypothesis to generate all kinds of combinations of permissions. But the pytest fixtures we use are generated once per test, and not once per hypothesis example.

So as is the PR solves the issue, and tests more than it did. More tests would just make it that much more complete.

mxmehl:
Thanks for the explanation.

@nicorikken would you like to have a look at this before merging it? My experience with tests is rather limited.

nicorikken:
I tested it and it works. Previously it would trigger an error:
```
usage: reuse addheader [-h] [--copyright COPYRIGHT] [--license LICENSE] [--year YEAR] [--style {applescript,aspx,bat,bibtex,c,css,f,haskell,html,jinja,jsx,lisp,m4,ml,plantuml,python,rst,tex}]
                       [--copyright-style {spdx,string,string-c,string-symbol,symbol}] [--template TEMPLATE] [--exclude-year] [--single-line] [--multi-line] [--explicit-license] [--skip-unrecognised]
                       path [path ...]
reuse addheader: error: argument path: can't write to 'test.sh'
```
Now it is successful:
```
Successfully changed header of test.sh.license
```

One thing to consider for future improvements, the `--explicit-license` is needed, otherwise you get presented with a stacktrace:
```
  File "/home/nico/Software/fsfe/reuse-tool/venv/bin/reuse", line 8, in <module>
    sys.exit(main())
  File "/home/nico/Software/fsfe/reuse-tool/venv/lib/python3.8/site-packages/reuse/_main.py", line 256, in main
    return parsed_args.func(parsed_args, project, out)
  File "/home/nico/Software/fsfe/reuse-tool/venv/lib/python3.8/site-packages/reuse/header.py", line 605, in run
    result += _add_header_to_file(
  File "/home/nico/Software/fsfe/reuse-tool/venv/lib/python3.8/site-packages/reuse/header.py", line 436, in _add_header_to_file
    with path.open("w", encoding="utf-8", newline=line_ending) as fp:
  File "/usr/lib/python3.8/pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/usr/lib/python3.8/pathlib.py", line 1078, in _opener
    return self._accessor.open(self, flags, mode)
PermissionError: [Errno 13] Permission denied: 'test.sh'
```
It seems to me like that is a case that needs a more elegant solution, but that can be solved outside of this PR.

Having checked the code, this looks good to merge!

Sorry @CharString for not getting to this sooner.

PR #349, opened 2021-04-22
Optionally determine copyright year via Git

### LINKED ISSUE #297: Automatically determine copyright dates
In using REUSE to add headers to existing projects, I find myself looking through the git history to determine the first year of contributions in order to determine the copyright span. It would be nice if REUSE was able to do this work automatically. I think of a behavior like:

```
$ reuse addheader --copyright "Nico Rikken" --license  GPL-3.0-or-later
No date specified, trying to determine it from version control.
Found contributions from 2018-2021, using it in the copyright information.
```

If this is not desired by default, it could also be triggered using something like a  `--detect-dates` flag.

Also more advanced scenario's call for determining history on a per-file basis, and dealing with intermittent contributions (`2010, 2014-2016, 2020`). The easiest is to consider the repo as a whole and take a complete date span, as demonstrated above.

rpavlik:
I was looking for a similar functionality as #297 and decided to just code it in to the tool instead of wrapping the python API in another tool. I've split the changes up into two: the first is just an extract-function refactor that is a prerequisite for doing per-file years without code duplication.

I'm using GitPython, though mostly just as a wrapper for finding git and calling it in a nice way, because I couldn't figure out how to get the log "natively" thru git-python, I'm pretty new to using it (this is only my second project using it). It's an optional dependency, the command-line argument "--use-first-commit-year" only shows up if it's loadable.

Right now, I have it set up so that if you pass "--use-first-commit-year", it will use whatever year is earliest, of the earliest author year on the file or the specified/defaulted-to-today year. (It does do --follow, but it checks every commit's date, in case something is out of order for some reason. A loop was cheap and easy and meant I didn't have to think about pipelines across platforms) So if you put the LICENSE year which might be the one from the start of the project, the argument is essentially a no-op since no commits were authored before then  (as happened to me when I was trying this on https://github.com/ValveSoftware/openvr ). I have this marked with TODO because I'm not sure which is the best option.

Fixes #297

### DIFF
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -27,6 +27,14 @@ from jinja2 import Environment, FileSystemLoader, PackageLoader, Template
 from jinja2.exceptions import TemplateNotFound
 from license_expression import ExpressionError
 
+try:
+    # pylint: disable=unused-import
+    import git
+
+    HAVE_GIT = True
+except ImportError:
+    HAVE_GIT = False
+
 from . import SpdxInfo
 from ._comment import (
     EXTENSION_COMMENT_STYLE_MAP,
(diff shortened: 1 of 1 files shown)


rpavlik:
Sorry for all the force-pushes, I couldn't repro some of these test failures locally and/or I got different test failures locally

mxmehl:
Thank you! I will make sure your PR is being reviewed soonish :)

nicorikken:
Thanks. I just did a quick review. I think it is elegant to take the earliest, of another statement already exists. So I think the TODO could be removed. Throwing a warning would be good though.
Also, your implementation is git-specific, but other parts of reuse-tool are also compatible with other vcs tools. Maybe the structure and naming can be kept universal, so other vcs tools could be added at a later stage.

rpavlik:
Thanks for the review. I didn't realize other parts of the tool were vcs generic, since there's an faq entry explaining why only Git is supported 😁 Tbh, the way I did that particular bit, with gitpython mostly just to find the repo and executable and to wrap the call, isn't ideal in my mind: I was hoping/thinking I could use gitpython entirely without shelling out to git. But, it did work for the task I had that day. I'd be more than ok if somebody else took this and finished/redid that little bit of vcs integration in a better way. I tried to keep the commits neatly separated into "refactoring needed as a prerequisite" and "the actual git stuff", thinking that even if the last integration wasn't quite up to standards, the earlier commits might be merged on their own or as a part of a revision 2 of the pull request.

Not sure when I'll get a chance/excuse to revisit this, but I'm keeping it in my GitHub notifications so it's on my list.

mxmehl:
Thanks again for this contribution, and sorry for coming so late with a decision.

Today we are having a hackathon during which we concluded that this VCS magic for grepping licensing/copyright info creates a lot of edge case issue. Please see https://github.com/fsfe/reuse-tool/issues/297#issuecomment-1019237349

However, we would be happy to provide this functionality, especially for batch operations, in an external helper script.

PR #353, opened 2021-04-29
feat: support capitalized .XML extension

nicorikken:
I encountered capitalized .XML extensions rather than the more typical lower
case .xml extensions. This commit adds that support.

Perhaps in the future we should add a property whether capitalization should be
considered or ignored.

### DIFF
--- a/src/reuse/_comment.py
+++ b/src/reuse/_comment.py
@@ -574,6 +574,7 @@ EXTENSION_COMMENT_STYLE_MAP = {
     ".tsx": JsxCommentStyle,
     ".vala": CCommentStyle,
     ".xml": HtmlCommentStyle,
+    ".XML": HtmlCommentStyle,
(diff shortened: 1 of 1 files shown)


mxmehl:
Oh, I wasn't aware that there is case-sensitivity with the extensions. Before adding an exception for XML, shouldn't we rather generally match case-insensitively for all extensions and also the special files like `.gitignore`? Or is there a downside I do not see?

nicorikken:
@mxmehl neither did I. Then again I was surprised to find an `.XML` extension rather than the `.xml` extension I would consider 'normal' :D Matching case-insensitively should work so long as there are no conflicts. But for matching specific filenames I'm a bit weary. A lot of applications expect files with specific filenames on specific locations. I'll just close this PR for now and I intend to open a separate issue where we can work on a proper solution.

PR #358, opened 2021-05-19
Add --rm to docker examples

Marzal:
Improve example so multiples executions don't generate a lot of containers that have no use anymore.

Great application by the way.


mxmehl:
Good idea, thanks!

PR #359, opened 2021-05-20
feat: ignore case when matching files

### LINKED ISSUE #355: Deal with capitalization
Sometimes you encounter filenames with an extension that is in a different capitalization. This was already the case for issues #353  and #306.

Some filetypes or filenames are only considered if they have an exact capitalization, so in that regard there is still value in having different handling if files have a different capitalization. Examples can be `Dockerfile`, `.gitignore` and `MANIFEST.in`.

Does the comment handling need to be expanded to have some capitalization parameters per file, or can we suffice with a broader policy?

nicorikken:
Changes the default behavior of the matching of filenames and extensions to be
case-insensitive. His is done by casting the COMMENT_STYLE_MAPS's and the files
to be matched to lowercase. I've chosen this runtime casting approach over
redefining the maps in lowercase to have the filenames and extensions in their
more common form, which can be capitalized like 'Dockerfile' or 'MANIFEST.in'.
If we need to get back on the capitalization behavior, we would still have the
definitions in a proper format.

Fixes #355


mxmehl:
Thank you, works great and looks like a clean solution!

May I ask you to remove the duplicated extension maps from _comment.py in the same run? These should be `F`, `F90`, and `ML`.

nicorikken:
Good catch, I added that.

Issue #230, opened 2020-05-14
addheader should not break XML files with <?xml ... ?> declaration


PhilippWendler:
Currently, `reuse addheader` treats XML files like HTML files and inserts a `<!-- comment -->` in the very beginning of the file. However, XML files may contain an [XML declaration](https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-prolog-dtd) of the form `<?xml ... ?>`, which if present must be the first thing in the file. Comments are only allowed after this declaration, not before. So calling `reuse addheader` on files with such a declaration makes them invalid XML.

I guess `addheader` could detect the XML declaration just like it [detects a shebang line](https://github.com/fsfe/reuse-tool/blob/014cc70ce3a2a86b5a984b343c4d3bcc701d7767/src/reuse/header.py#L250) and insert the comment after it. I would even suggest to do this independently of the file extension, because every file that does contain a `<?xml ... ?>` declaration at the beginning is certainly expected to be valid XML by somebody and should be handled accordingly.

mxmehl:
Thanks for reporting this bug! And sorry for hijacking the thread, but the tool should perhaps make it easier to declare such declarations.

For instance, [some YAML files](https://github.com/fsfe/reuse-website/blob/master/.drone.yml) have to start with the first line being `---`, and addheader does not respect that either. And I guess, shebang, XML declarations and YAML will not be the only conditions the tool would have to detect.

carmenbianca:
Some Markdown files also begin the first couple of lines with something like:

```markdown
---
title: My Title
author: Me

# This is a comment in yaml/toml/whatever inside of a Markdown file
---

# Chapter 1

Hello, world.

<!-- This is a comment inside of Markdown -->
```

That one is a lot harder to get right.

#211 is effectively the same issue, which I had closed wontfix. Giving the user the tools to solve this would be nice, as @mxmehl proposed. One static solution is templates, but that mightn't work for all use-cases.

Not sure what the right thing to do here is.

PhilippWendler:
Indeed, such cases seem more tricky. `---` might also be too common to simply treat it as "never put our header in front of it" like it is done with `#!` and probably possible with `<?xml`.

So maybe fix the clear case `<?xml` first and then work towards a more general solution?

mxmehl:
Would it be a lot of work to have a regex per file type? So all files ending with `.sh` would take care of shebangs, and all files ending with `.yaml` or `.yml` would respect the `---`.

[...]

PR #378, opened 2021-06-13
Do not break XML files special first line

hoijui:
The special first-line in XML file needs to remain first line, else, the file becomes invalid regarding in the XML spec, and most editors will not recognize the file.
This PR uses the same technique used for shebangs, to protect the first lien in XML files.

This solves #230 (barely).


nicorikken:
Clever reuse of the existing code.

mxmehl:
That looks like a more generally usable and extensible solution, thank you!

PR #386, opened 2021-06-18
Add support for AsciiDoc

lfrancke:
https://docs.asciidoctor.org/asciidoc/latest/comments/

I'm afraid the C comment style only matches partially (multi line is different) but I'm not sure when multi line is chosen. I can also adapt to add a new style.

I chose the extensions that are listed here: https://asciidoctor.org/docs/asciidoc-recommended-practices/#document-extension


mxmehl:
Thank you! Multi-line is different indeed, but for `addheader` it should suffice as we only add it in single lines.

PR #393, opened 2021-07-06
docs: update package reference for Arch Linux

grawlinson:
The reuse package has been moved to the community repository from the AUR. 👍


mxmehl:
Oh wait what, really? I will not complain as this is one maintenance job less, but it's kinda surprising :)

mxmehl:
Just saw the emails to my AUR address, thanks for taking over maintenanceship! Great to know that it's supported officially now 👍

PR #397, opened 2021-08-03
Vue.js single file component header support

### LINKED ISSUE #395: vue file support for addheader
`.vue` files should be supported.

[The format is kind of similar to HTML](https://v3.vuejs.org/guide/single-file-component.html#what-about-separation-of-concerns)

weeman1337:
Vue.js single file components are using the HTML comment style, [see Vue.js doc](https://v3.vuejs.org/guide/single-file-component.html#what-about-separation-of-concerns)

closes #395


mxmehl:
Thanks! Just added the changelog entry right away

PR #407, opened 2021-09-06
Use C-style comments for React JSX and TSX files

### LINKED ISSUE #406: React JSX / TSX comment style should use // prefix
Currently, `.jsx` and `.tsx` files get this kind of header:

```js
{/*
SPDX-FileCopyrightText: 2021 Somebody
SPDX-License-Identifier: GPL-3.0
*/}
```

However, this style of commenting is not supposed to be used for top-level comments within the Javascript code, but rather for embedded HTML within components, e.g.:

```
export const Card = ({ title, paragraph }: CardProps) => <aside>
  <h2>{ title }</h2>
  <p>
    {/* Wow look, it's a paragraph! */}
    { paragraph }
  </p>
</aside>
```

Unfortunately, when the REUSE header is run through [Prettier](https://prettier.io/) (which is a very popular tool these days, for good reasons), it gets changed to:

```js
{
  /*
SPDX-FileCopyrightText: 2021 Somebody
SPDX-License-Identifier: GPL-3.0
*/
}
```

which is ugly and unnecessarily verbose.  This comes from the following code: 

https://github.com/fsfe/reuse-tool/blob/d546b1449aea93ac1bffe677e430e1bc004c8435/src/reuse/_comment.py#L363-L368

I suggest that we simply change it so that `*.jsx` and `*.tsx` use `CCommentStyle` instead:

https://github.com/fsfe/reuse-tool/blob/d546b1449aea93ac1bffe677e430e1bc004c8435/src/reuse/_comment.py#L281-L291

aspiers:
The `{/* ... */}` comment style is intended for use when embedding comments within HTML within JSX and TSX, and not for top-level comments.  So switch to C-style comments which works fine with both JSX and TSX.

Fixes #406.


mxmehl:
Strange, when I read [this for instance](https://wesbos.com/react-jsx-comments), it seems it's not that simple:

> Kind of a pain, but React devs has stated they do not plan to add regular JS commenting to JSX.

Could you please confirm that C-style comments are correct for top-level comments?

mxmehl:
Actually, if you want to use an existing style for a file extension, you can just change it here to `CCommentStyle`:

https://github.com/fsfe/reuse-tool/blob/d546b1449aea93ac1bffe677e430e1bc004c8435/src/reuse/_comment.py#L524

aspiers:
[**@mxmehl**](https://github.com/mxmehl) commented on [September 17, 2021 4:05 PM](https://github.com/fsfe/reuse-tool/pull/407#issuecomment-921870995 "2021-09-17T15:05:59Z - Replied by Github Reply Comments"):
> Actually, if you want to use an existing style for a file extension, you can just change it here to `CCommentStyle`:
> 
> https://github.com/fsfe/reuse-tool/blob/d546b1449aea93ac1bffe677e430e1bc004c8435/src/reuse/_comment.py#L524

I did consider that but it seemed like removing a class would break backwards-compatibility; not sure if anyone is relying on the API though?

aspiers:
[**@mxmehl**](https://github.com/mxmehl) commented on [September 17, 2021 4:03 PM](https://github.com/fsfe/reuse-tool/pull/407#issuecomment-921869064 "2021-09-17T15:03:30Z - Replied by Github Reply Comments"):
> Strange, when I read [this for instance](https://wesbos.com/react-jsx-comments), it seems it's not that simple:
> 
> > Kind of a pain, but React devs has stated they do not plan to add regular JS commenting to JSX.

That is talking specifically about comments inside HTML blocks, not outside.  AFAIK everything outside is just regular JS.  (Also note that article is from 2015.)  Even the official React docs say the same thing, focusing purely on comments inside the HTML:

https://reactjs.org/docs/faq-build.html#how-can-i-write-comments-in-jsx

> Could you please confirm that C-style comments are correct for top-level comments?

I believe so, because my understanding is that everything outside the JSX-specific HTML blocks is regular JS.  I can't find a completely independent and formal JSX language spec, which I believe is because JSX completely piggy-backs on top of the JS language, so it doesn't need a complete spec of its own.  This seems to be backed up by the following:

- https://facebook.github.io/jsx/
- https://uk.reactjs.org/blog/2014/09/03/introducing-the-jsx-specification.html
- https://en.wikipedia.org/wiki/JSX_(JavaScript)

which all describe JSX as an *extension* to JS.

Therefore it seems very clear to me that the `{/* ... */}` syntax is specifically for use within JSX HTML blocks, because it relies on the interpolation given by `{...}`, which makes no sense when used outside those blocks in normal JS context.

mxmehl:
> I did consider that but it seemed like removing a class would break backwards-compatibility; not sure if anyone is relying on the API though?

Good point. Could you please add a comment in the Changelog under "Breaking", explicitely mentioning the sole effect on the API?

> Therefore it seems very clear to me that the {/* ... */} syntax is specifically for use within JSX HTML blocks, because it relies on the interpolation given by {...}, which makes no sense when used outside those blocks in normal JS context.

OK, and if you confirm that linters and other tools react positively on this, I'd be fine with it. With the removal of the empty comment style and the Changelog entry (both in Changed and Breaking), I'd be happy to merge this.

janderssonse:
> [**@mxmehl**](https://github.com/mxmehl) commented on [September 17, 2021 4:03 PM](https://github.com/fsfe/reuse-tool/pull/407#issuecomment-921869064):
> 
> > Strange, when I read [this for instance](https://wesbos.com/react-jsx-comments), it seems it's not that simple:
> > > Kind of a pain, but React devs has stated they do not plan to add regular JS commenting to JSX.
> 
> That is talking specifically about comments inside HTML blocks, not outside. AFAIK everything outside is just regular JS. (Also note that article is from 2015.) Even the official React docs say the same thing, focusing purely on comments inside the HTML:
> 
> https://reactjs.org/docs/faq-build.html#how-can-i-write-comments-in-jsx
> 
> > Could you please confirm that C-style comments are correct for top-level comments?
> 
> I believe so, because my understanding is that everything outside the JSX-specific HTML blocks is regular JS. I can't find a completely independent and formal JSX language spec, which I believe is because JSX completely piggy-backs on top of the JS language, so it doesn't need a complete spec of its own. This seems to be backed up by the following:
> 
>     * https://facebook.github.io/jsx/
> 
>     * https://uk.reactjs.org/blog/2014/09/03/introducing-the-jsx-specification.html
> 
>     * https://en.wikipedia.org/wiki/JSX_(JavaScript)
> 
> 
> which all describe JSX as an _extension_ to JS.
> 
> Therefore it seems very clear to me that the `{/* ... */}` syntax is specifically for use within JSX HTML blocks, because it relies on the interpolation given by `{...}`, which makes no sense when used outside those blocks in normal JS context.

I would agree, I just stumbled upon this "problem" when adding headers in an electron app code base,  npm run lint works fine with top level comments of C-style as proposed ie. /* a multine top spdx reuse comment */, but gave multipline linting errors with the current reuse default style "{/* inline jsx comment style /*}". As @aspiers also summarizes I can only the current jsx comment style is only used for inline jsx blocks, (I also tried to google information about this and came to the same conclusion) and that per default the proposed comment style from @aspiers should be the one to use. I vote for merging this pr.

linozen (approved):
This seems straightforward to me. I don't  expect licensing information to ever be _inside_ an JSX block, so C-Style comments are probably more robust.

floriansnow:
> [**@mxmehl**](https://github.com/mxmehl) commented on [September 17, 2021 4:05 PM](https://github.com/fsfe/reuse-tool/pull/407#issuecomment-921870995):
> 
> > Actually, if you want to use an existing style for a file extension, you can just change it here to `CCommentStyle`:
> > https://github.com/fsfe/reuse-tool/blob/d546b1449aea93ac1bffe677e430e1bc004c8435/src/reuse/_comment.py#L524
> 
> I did consider that but it seemed like removing a class would break backwards-compatibility; not sure if anyone is relying on the API though?

I think from version 1.0.0 onwards, this will be easier, but we're not there yet.  :)  The problem is that changing the internal behavior of the class is just a different kind of breaking change that someone might rely on.  Perhaps @carmenbianca could pitch in here because you have more experience in how we treat backwards compatibility here.

janderssonse:
I'm just chipping in from an end user perspective here. For 0.x.x versions backwards compatibility would not be expected from a project, but it would be very nice if such a change is clearly documented. Semver spec also says: _Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable._  https://semver.org/. So as an end user I expect breaking changes in a 0.x.x versions, but I would also hopefully be notified about them in the changelog or alike.

floriansnow:
@janderssonse I agree with you, but not being on 1.x.x yet is a bit of an esoterical question in this case and we've been pretty careful about breaking changes.  From version 1 forward, we plan on using semantic versioning and that will facilitate communicating what kinds of changes a new version brings.  But let's see what Carmen thinks about this because she knows the past practice best.

linozen:
Just out of curiousity - what is the breakage we expect to see here if we migrate to C-style comments for `.jsx` files? In the end, `.jsx` is simply a "syntax extension" [[source](https://reactjs.org/docs/introducing-jsx.html)] to normal JavaScript files. It is very, very unlikely that the licensing info of a `*.jsx` file lives _inside_ a JSX block which usually don't start right away in any given file. They live at the top of the file, where normal C-style comments are perfectly fine. So, IMHO merging this as soon as possible and simply recording the change in the changelog would be perfectly fine.

mxmehl (approved):

floriansnow:
Ok, since there has not been any feedback on past practice in regards to breaking changes, I would suggest we move this ahead by
- removing the class `JsxCommentStyle`
- using `CCommentStyle` in the same place
- adding a note about a breaking change in the changelog
The breaking change should be rather small anyway IMHO and would most likely not affect anyone.

carmenbianca:
@floriansnow I agree with this approach. Feel free to do it like that.

linozen:
LGTM :+1:

PR #413, opened 2021-09-10
A few corrections

criadoperez:
A few translations that needed corrections.


mxmehl:
Thanks for your attention to detail!

Issue #364, opened 2021-06-09
Suggest similarly named licenses if BAD licnese detected


hoijui:
When a BAD license is detected, suggest similarly named, valid licenses, if any. For example:

* BAD license: `CC0`
* Suggest: `CC0-1.0`

hoijui:
Later I noticed, That there are suggestions like that (I got one when trying a "BSD<something>" license).
I guess it could be improved still - see my example above: `CC0` does not result in a suggestion.

nicorikken:
I agree, similarly there is the distinction between `GPL-3.0` and `GPL-3.0-or-later`. If you naively type `GPL-3.0` you might be registering the wrong license. Considering this is all user-input, some validation would be nice. I'm pretty sure there is some Python fuzzy text matching library available to find similar identifiers to ask the user to confirm.

siiptuo:
This was implemented in #152. Originally the suggestions were better but adding a dependency for [the fuzzy string matching library](https://pypi.org/project/jaro-winkler/) was not possible, so the functionality was simplified. Currently the suggestions work well only on typos such as `GLP-3.0-ro-ltaer`.

I agree that the suggestion should be improved. Is there any fuzzy string matching library we could use? Otherwise, the current solution could be modified to match prefixes so that inputting `GPL-3.0` would suggest `GPL-3.0-only` and `GPL-3.0-or-later` for instance.

mxmehl:
Oh, I almost forgot the feature you've once added, @siiptuo. Wouldn't that already be a great start for a check in `addheader`, especially after the merge of #416?

mxmehl:
I renamed this issue to track the effort of adding the functionality that we've had since #416 also for `addheader`. Currently there is no sanity check at all.

hoijui:
still relevant

... though now I don;t even know how I tested that, or where the suggestions were shown. `reuse lint` does not show me a suggestion anymore, neither for `BSD` nor for `GLP-3.0-ro-ltaer` nor `CC0`.

PR #416, opened 2021-09-21
Suggest valid SPDX identifiers based on prefix

siiptuo:
Related to #364. Entering invalid identifiers like `CC0` and `GPL` will now suggest valid identifiers which start with the inputted string while typos are still detected as before.


mxmehl:
Tested locally and I like the simple solution! Just reworded the test description a bit, and added a changelog entry.

PR #418, opened 2021-09-29
Fix regression for files without permissions.

CharString:
Closes #398

PR #421, opened 2021-10-01
Add isort to pre commit hooks

CharString:
When working on #418, one of my commits tripped on isort in CI. This should prevent this from happening in the future.


mxmehl:
Good catch, thank you.

Issue #420, opened 2021-10-01
Add auto-detection of comment styles for more JavaScript-related file types


silverhook:
I identified the following formats that can be easily associated with existing comment styles:

`.yarnrc` file – `PythonCommentStyle`
`.npmignore` file – `PythonCommentsStyle` ([same as `gitignore`](https://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html))
`.eslintignore` file – `PythonCommentsStyle` ([same as `gitignore`](https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file))
`.eslintrc` file – `UncommentableCommentStyle`
`yarn.lock` file  – `UncommentableCommentStyle`(simply because it is [auto-generated](https://classic.yarnpkg.com/en/docs/yarn-lock/))

`*.soy` extension – `CCommentStyle` (same as JS)

`*.tpl` extension – [needs a custom style](https://stackoverflow.com/a/4058880):

```
 MULTI_LINE = ("{*", "", "*}")
```

Happy to provide a pull request, if no-one else wants to tackle it.

PR #428, opened 2021-10-19
Add comment styles for more JS-related file types

silverhook:
fixes https://github.com/fsfe/reuse-tool/issues/420

Please do not implement yet. While this patch should work, I am still exploring whether there is a better way to handle `.tpl` files.


silverhook:
Fixed. Sorry for the messy merge request.

PR #439, opened 2021-11-30
Add CC-BY-ND-3.0-DE as supported license

Darneas:
SPDX already supports this license.


mxmehl:
Sorry for the wait. Replied to this in #441. The process would be a bit different, and just makes sense before a release usually.

However, would it help you if we had this change just on the main/master branch, or would you need a release?

mxmehl:
Just released 0.14.0 with the 3.15 license list included. That should also contain the desired license :)

PR #446, opened 2021-12-25
Add support for meson_options.txt

### LINKED ISSUE #445: Add --style for meson files
The meson build system[1] uses `meson.build` and `meson_options.txt`. Both use
`#` for comments.

[1] https://mesonbuild.com/

nirs:
We added support for meson.build file in #251 but we need also to
support meson_options.txt[1], used to add build options.

[1] https://mesonbuild.com/Build-options.html

Fixes #445.


mxmehl:
Thanks for this as well :)

PR #448, opened 2021-12-25
Support Containerfile

### LINKED ISSUE #447: Support Containerfile
Podman[1] supports now both Containerfile[2] and Dockerfile. This is another name for
the same file format. 

[1] https://podman.io/
[2] https://docs.podman.io/en/latest/markdown/podman-build.1.html

nirs:
The podman tool supports now both Containerfile and Dockerfile. This is
another name for the same file format.

Fixes #447.


mxmehl:
Thank you!

Issue #454, opened 2022-01-07
utf-8 in dep5 file results in errors on windows


thbde:
The following description can be reproduced via: https://github.com/thbde/reuse-utf8-dep5-issue

Assume that we use the dep5 file to declare the license state for a repository.
Furthermore, the dep5 file contains utf-8 characters (or code points).

In that case, reuse will fail if we execute:
`reuse download --all`

And the reported errors are rather confusing:
```
# Windows 10 64 bit
$ python --version
Python 3.10.1
$ reuse --version
reuse 0.14.0
$ python -c 'import locale;print(f"{locale.getpreferredencoding()=}")'
locale.getpreferredencoding()='cp1252'

$ reuse download --all
reuse.report - ERROR - Could not read 'call_open.py'
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 321: character maps to <undefined>
reuse.report - ERROR - Could not read 'file1.txt'
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 321: character maps to <undefined>
reuse.report - ERROR - Could not read 'file2.txt'
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 321: character maps to <undefined>
reuse.report - ERROR - Could not read 'file3.txt'
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 321: character maps to <undefined>
reuse.report - ERROR - Could not read 'output_reuse_call.txt'
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 321: character maps to <undefined>
```

The errors want to tell us:

1. There is a UnicodeDecodeError in each file. But none of these files uses any special unicode or utf-8, all are ascii (only).
2. The error is at the exact same place for all files. How can this be?

The root cause (after quite some debugging) is the open call for the dep5 file:
https://github.com/fsfe/reuse-tool/blob/60c0986bb24a3b482ee0527e9195f7c23cadb003/src/reuse/project.py#L216-L219

1. The file is opened via open() which uses a platform dependent encoding as a default (`locale.getpreferredencoding()` which is `cp1252` on windows)
2. The `Copyright` class does somewhere deep inside quite some nested functions a
```
for element in fd:
  ...
```
(more precisely, here: https://salsa.debian.org/python-debian-team/python-debian/-/blob/278e016f5ed8d3ed4fa17d7c30b54c149d428808/lib/debian/deb822.py#L759 )

[...]

PR #455, opened 2022-01-07
add encoding to open() calls

thbde:
This PR contains two commits.

https://github.com/fsfe/reuse-tool/commit/94226e9d4d836ab3770483627bc76580442ad373 fixes the issue described in https://github.com/fsfe/reuse-tool/issues/454

https://github.com/fsfe/reuse-tool/commit/f168ccd3dfec313f9abb183bf92f893a7e007509 adds encoding to all `open()` that I found via grep (ignoring 2 cases that are definitely not relevant).

It may also be good to add tests for that issue, but adding such tests is not trivial because we need to mock and manipulate some system specific stuff. I do not know the code base of the reuse tool well enough to implement this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,12 +45,12 @@ fallback_version = "0.14.0"
 
 def readme_md():
     """Return contents of README.md"""
-    return open("README.md").read()
+    return open("README.md", encoding="utf-8").read()
 
 
 def changelog_md():
     """Return contents of CHANGELOG.md"""
-    return open("CHANGELOG.md").read()
+    return open("CHANGELOG.md", encoding="utf-8").read()
 
 
 class BuildTrans(cmd.Command):
--- a/src/reuse/_licenses.py
+++ b/src/reuse/_licenses.py
@@ -24,7 +24,7 @@ def _load_license_list(file_name):
     from https://github.com/spdx/license-list-data
     """
     licenses_map = {}
-    with open(file_name, "r") as lics:
+    with open(file_name, "r", encoding="utf-8") as lics:
         licenses = json.load(lics)
         version = licenses["licenseListVersion"].split(".")
         for lic in licenses["licenses"]:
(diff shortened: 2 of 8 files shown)


mxmehl:
Thank you for your great issue and the PR.

I am not familiar with Windows systems. Would hardcoding the encoding to UTF-8 conflict with files using other encodings?

carmenbianca:
> Would hardcoding the encoding to UTF-8 conflict with files using other encodings?

UTF-8 is technically already hardcoded. reuse already implicitly assumes UTF-8 encoding. I think there's some code somewhere that 100% definitely assumes UTF-8, but `open(file)` without `encoding='...'` defaults to UTF-8 on Linux, which is functionally identical to hardcoding it. [Documentation here](https://docs.python.org/3/library/functions.html#open)

> In text mode, if encoding is not specified the encoding used is platform-dependent: locale.getpreferredencoding(False) is called to get the current locale encoding.

mxmehl:
OK, so no problems to expect on machines other than Linux, or when editing files that are not UTF-8 encoded?

carmenbianca:
> OK, so no problems to expect on machines other than Linux, or when editing files that are not UTF-8 encoded?

Well…

reuse already naively assumes UTF-8 here and there in the code. This PR just puts that assumption _everywhere_ in the code base.

Whether that's a good idea, I don't know. I remember that a very early of this tool used a library to detect encoding (or rather, it was used to detect binaries, and encoding detection was a nifty extra), but it was very slow.

But! So far there have been no reported issues on the UTF-8 assumption in reuse, so it's probably fine!

Additionally, this fixes a problem in #484 ;)

mxmehl:
Right, and I see we came to the same conclusion in #221.

However, quite some tests failed.

carmenbianca:
In the process of merging this in #491.

carmenbianca:
Closed by #491. Thank you!

PR #457, opened 2022-01-19
Recommend installation via pipx

hexagonrecursion:
Managing python, pip and virtual environments can be too much work sometimes especially for a project that is not itself written in python. pipx is a convenient shortcut, but people unfamiliar with the python ecosystem might not even know it exists. To mitigate this I have added an example of running your tool via pipx to your readme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ The versions follow [semantic versioning](https://semver.org).
 
 ### Added
 
+- Recommendations for installation/run methods: package managers and pipx (#457)
+
 ### Changed
 
 - Use `setuptools` instead of the deprecated `distutils` which will be removed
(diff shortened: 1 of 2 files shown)


mxmehl:
Thank you! I never heard of pipx before, but it looks good on first sight.

However, I would like to keep the package managers on top of the installation instruction list. They are easiest and safest to use for most users, and keep track of updates (in most cases). But for me it would be fine to rank pipx second if @carmenbianca agrees.

carmenbianca:
pipx is great. I'd almost be tempted to remove the 'installation via pip' instructions, but I think they're valuable to keep around.

Putting pipx second is a good idea. To push users towards pipx instead of pip, maybe append '(Recommended)' to both the distro instructions and pipx?

mxmehl:
Thank you, just merged after adding a changelog entry and a cosmetic change.

PR #458, opened 2022-01-21
Add prettier to the project

carmenbianca:
Life gets a lot easier when you let auto-formatters do all the formatting for you. This is a simple QOL addition.

### DIFF
--- a/.github/workflows/prettier.yaml
+++ b/.github/workflows/prettier.yaml
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: 2022 Carmen Bianca Bakker <carmenbianca@fsfe.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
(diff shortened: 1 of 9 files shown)


mxmehl:
What's the workflow doing? Is it checking whether prettier runs successfully, or is it actually doing changes on the code that last?

Also, I am not so happy with running Github actions by quite random people (no offense). Especially as we do not want to get too dependent on Github and want to keep a migration to a Free Software forge easy.

carmenbianca:
@mxmehl 

> Is it checking whether prettier runs successfully, or is it actually doing changes on the code that last?

Both, I think, although it doesn't end up committing anything. In any case you're right; it should probably only check.

> Also, I am not so happy with running Github actions by quite random people

Agreed. I can re-jig that quite easily I think.

mxmehl (approved):

PR #462, opened 2022-01-22
Rename Mary Sue to Jane Doe

### LINKED ISSUE #460: Rename Mary Sue to Jane Doe
I'm not super happy with Mary Sue as the example person in a lot of tests. At the time I wrote this, 'Mary Sue' was a bit of a running gag in a subculture I was in, understood as [the Wikipedia definition](https://en.wikipedia.org/wiki/Mary_Sue):

> A Mary Sue is a type of fictional character, usually a young woman, who is portrayed as unrealistically free of weaknesses. Originating in fan fiction, a Mary Sue is often an author's idealized self-insertion.

Since then, or maybe even at the time, people have begun to use the name as almost synonymous to 'any (strong) female character I don't like'. That leaves a sour taste in my mouth, and I'd rather just rename Mary Sue in this code base.

Gri-ffin:
I replaced all occurrences of Mary Sue to Jane Doe in all test files, I also found Mary Sue in the header of some source files,
but I replaced them due to you mentioning that you wanted to replace the name in all code base.

fixes #460

### DIFF
--- a/tests/test_header.py
+++ b/tests/test_header.py
@@ -24,11 +24,11 @@ from reuse.header import (
 def test_create_header_simple():
     """Create a super simple header."""
     spdx_info = SpdxInfo(
-        {"GPL-3.0-or-later"}, {"SPDX" "-FileCopyrightText: Mary Sue"}
+        {"GPL-3.0-or-later"}, {"SPDX" "-FileCopyrightText: Jane Doe"}
     )
     expected = cleandoc(
         """
-        # spdx-FileCopyrightText: Mary Sue
+        # spdx-FileCopyrightText: Jane Doe
         #
         # spdx-License-Identifier: GPL-3.0-or-later
         """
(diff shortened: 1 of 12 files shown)


tests/test_header.py
```diff
@@ -298,7 +298,7 @@ def test_find_and_replace_keep_shebang():
         #!/usr/bin/env python3
 
         # spdx-FileCopyrightText: Jane Doe
-        # spdx-FileCopyrightText: Mary Sue
+        # spdx-FileCopyrightText: Jane Doe
```
carmenbianca, line 301: There are two Jane Does here :) One of them should probably be the kindly John Doe.

carmenbianca (changes requested):
Thank you @Gri-ffin for your PR! I've made a small request for a change in this review. Feel free to add yourself to AUTHORS.rst if you want.

tests/test_header.py
Gri-ffin: Ok, I can fix that

carmenbianca (approved):
LGTM. Thank you very much for this contribution @Gri-ffin !

PR #464, opened 2022-01-22
Ignore REUSE ignore blocks

### LINKED ISSUE #463: Introduce marking a block that lint ignores
Users should be able to mark a block that `reuse lint` ignores. This would solve the problem that users experience when they use the word "copyright" or even the strings "SPDX-License-Identifier" or "SPDX-FileCopyrightText" somewhere in the file without actually meaning to provide REUSE-related information.

Simplified example from #70:

```sh
#!/usr/bin/env bash
# SPDX-FileCopyrightText: 2021 John Doe
#
# SPDX-License-Identifier: CC0-1.0

echo "SPDX-FileCopyrightText: $(date +'%Y') Jane Doe" > file.txt
echo "SPDX-License-Identifier: MIT" > file.txt

exit 0
```

This would lead the the following error when running `lint`:

```
$ reuse lint
reuse._util - ERROR - Could not parse 'MIT" > file.txt'
reuse.project - ERROR - 'foobar.sh' holds an SPDX expression that cannot be parsed, skipping the file
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
* foobar.sh
[...]
```

In a sprint we had the idea to define a begin/end string that marks blocks that are to be ignored by `lint`. E.g.:

```sh
#!/usr/bin/env bash
# SPDX-FileCopyrightText: 2021 John Doe
#
# SPDX-License-Identifier: CC0-1.0

# REUSE-IGNORE-BEGIN
echo "SPDX-FileCopyrightText: $(date +'%Y') Jane Doe" > file.txt
echo "SPDX-License-Identifier: MIT" > file.txt
# REUSE-IGNORE-END

exit 0
```

## To discuss

- [x] Define the exact Begin/End strings. Check whether SPDX provides them
- [x] Does it require a specification change?

carmenbianca:
Fixes #463


src/reuse/_util.py
```diff
@@ -211,13 +211,19 @@ def filter_ignore_block(text: str) -> str:
     REUSE_IGNORE_END to remove lines that should not be treated as copyright and
     licensing information.
     """
+    reuse_ignore_begin = "REUSE_IGNORE_BEGIN"
+    reuse_ignore_end = "REUSE_IGNORE_END"
```
carmenbianca, line 215: These should probably be global variables :)

floriansnow: @carmenbianca Yeah, I was wondering about that too.  Feel free to put them in an appropriate place.

carmenbianca:
- Add tests for REUSE_IGNORE_END missing
-  Add tests for both of them on same line
- Add test for reversed things

nicorikken:
After this merge we can rework the codebase to remove the 'hacks' like rewrites and concetations that are used to avoid writing correct SPDX headers.

src/reuse/_util.py
floriansnow: Done.

carmenbianca:
TODO:

- change log
- rewrite history
- figure out whether 'ignore block' is actually correct nomenclature
- await @mxmehl 's analyses on what the tags' names should actually be

floriansnow:
>     * rewrite history

What exactly do you expect here?  I actually considered this a clean history state.  :-)


>     * figure out whether 'ignore block' is actually correct nomenclature
>     * await @mxmehl 's analyses on what the tags' names should actually be

I guess those two are the same?  Or is there anything else I can do in that respect?

carmenbianca:
> What exactly do you expect here? I actually considered this a clean history state. :-)

Ah, I'm fairly used to bundling features into a small number of commits just before merging a PR. But honestly the history as-is is perfectly serviceable :)

> I guess those two are the same?

Almost the same :) I want to know whether 'ignore block' is the correct term for the thing we're writing here, or whether there exists a different term used by other programs. The exact names of the tags, then, are an implementation detail of the 'ignore block' functionality.

mxmehl:
> Almost the same :) I want to know whether 'ignore block' is the correct term for the thing we're writing here, or whether there exists a different term used by other programs. The exact names of the tags, then, are an implementation detail of the 'ignore block' functionality.

I wasn't able to find ignore blocks in the SPDX spec, but changed my mind anyway on the "ignore blocks" name.

We should be careful with using the word "ignore" in the scope of REUSE. REUSE only ignores [a handful of files](https://reuse.software/spec/#copyright-and-licensing-information) for very good reasons: apart from files that are ignored by the VCS only files in `.reuse` and the license text files. Ignoring freely definable block in files would seem like you could hide rightfully licensed stuff by third parties. 

Instead, I propose something along the lines of `REUSE-LintEscape-Begin`. The mention of "lint" makes clear that it's not about ignoring parts for REUSE in general, but for the purely technical linting process. "escape" also feels a bit more cautious and like an active decision than "ignore". I am not fully convinced of the term yet, but perhaps you understand where it's headed to.

Also, I want to raise the question of capslock and undersore/dash. I propose to stick with how people know it from SPDX: either dash or CamelCase, but apart from REUSE/SPDX only capital first letters (exception CamelCase of course).

nicorikken:
I now think such an ignore block should actually be defined generally in say SPDX, because many different tools can encounter the exact situation. It makes no sense to only ignore it for reuse-tool.

@mxmehl Regarding the naming, I don't think that 'lint' is the right term, it reminds me of code quality. I think main issue you express in the solution is more about the generic naming. 'REUSE_IGNORE' doesn't explain the intent. Rather than changing the term to 'lint' I think we can be more expressive about _why_ you need to ignore:

1. `SPDX-IgnoreUnintendedCopyright`
2. `SPDX-IgnoreFalsePositive`
3. `SPDX-IgnoreInvalidCopyright`
4. `SPDX-IgnoreMeaninglessCopyright`

And similar for license details, like `SPDX-IgnoreMeaninglessLicense`, all to be as specific as possible.

As far as I know the SPDX standard, they don't have the concept of blocks, but rather a ByteRange like a SnippetByteRange. I quite dislike that approach, but that would be something to discuss in the SPDX context.

For now it might still make more sense to keep it reuse-specific:

```
REUSE-IgnoreMeaninglessCopyrightStart
testcode = "Copyright Jane Doe 2020"
REUSE-IgnoreMeaninglessCopyrightEnd
```

Regardless of the exact term, I think we should express the need for specific ignore options in the SPDX project, hoping that eventually a more generic term will be adopted in the standard.

mxmehl:
Valid arguments, thank you Nico.

As you said, SPDX is mostly thinking in ranges now. But with the snippets, I proposed a begin/end tag already, see spdx/spdx-spec#464. So it may be possible to also get the ignore tag into SPDX, but I have a number of concerns:

1. For SPDX, false-positives are a non-issue mostly, at least currently. It is obvious that `echo "SPDX-License-Identifier: $license"` is not a SPDX statement. However, I think that some other downstream projects face the same issue, e.g. Scancode or FOSSology. @pombredanne @mcjaeger, what do you think? If more projects than just REUSE raise the need, chances might be better.
2. We are already blocked by SPDX for REUSE.yaml and the snippets, and I wouldn't like to also depend on SPDX 3.0 for this important feature that would solve a number of bad hacks. Therefore, I like your proposal of starting with a REUSE-specific tag anyway.

Now, to the tag: I saw that you refer on most examples to copyright, but license-identifiers have the same issue. Therefore, such a tag should be generic enough. I specifically liked the "FalsePositive" term because this is what it is.

What about one of the following:

* `REUSE-PotentialFalsePositiveBegin`
* `REUSE-FalsePositiveEscapeBegin`
* `REUSE-EscapeTagsBegin`
* `REUSE-EscapeExpressionsBegin`

nicorikken:
@mxmehl From your suggestions I like `REUSE-FalsePositiveEscapeBegin`. I dislike the term 'potential'.

mxmehl:
Why not "EscapeTags" or "EscapeExpressions" if I may ask? Personally I'd find them more understandable ;)

nicorikken:
Because I got hung up on the 'FalsePositive' and did not link 'tag' or 'expression' with SPDX so much. Now that you mention it I can see it being more on-point.

pombredanne:
@mxmehl  re: https://github.com/fsfe/reuse-tool/pull/464#issuecomment-1029884677

> 1. For SPDX, false-positives are a non-issue mostly, at least currently. It is obvious that `echo "SPDX-License-Identifier: $license"` is not a SPDX statement. However, I think that some other downstream projects face the same issue, e.g. Scancode or FOSSology. @pombredanne @mcjaeger, what do you think? If more projects than just REUSE raise the need, chances might be better.

Hey Max! Nice chattin's at FOSDEM!

Is the intent to have a way to tag text or code so it is ignored? but ignored from what? I am not sure I can ever see a use case for this. Ignoring a whole file sounds OK, but going inside a file does not seem reasonable to me or something that I would ever want a code author to have to worry about.

Or may be you are asking if this would be a problem in ScanCode or how we deal with possible things that could be in general false positives like license tags or identifiers seen in license detection tools? This is not a problem because  we deal with them by detecting them as false positive as needed either using license detection rules tagged as false positive or code to detect and uncover patterns of false positives.

I hope this helps!

mcjaeger:
hi, @mxmehl thanks for the ping!  there should not be too much of a meaning being put the escape tag naming. "invalid copyright", "unintended copyright" opens the door to the other questions, such as why it was unintended if someone wrote it there or what makes a copyright invalid? (as Philippe indicated above) Developers will argue about which tag is to be used exactly in one case and so on and so on.

Using "Potential" ... well I am not sure about that, because if it says "potential" you would like to capture it for scanning then. So there appears to be little benefit for using anything like "potential".

Having a tag which expresses: if you re going to pull that in the list of findings, it will be a false positive, appears sufficient. Other than that, would there be any logic that collects content of different escape blocks and list them separately in addition? If not, there is little benefit of using different tags for different sub cases.

As for the practical application, indeed scanning the fossology code itself brings the anbove mentioned issues and if the codebase could be augmented with "ignore that block". 

about 
>For SPDX, false-positives are a non-issue mostly, at least currently. It is obvious that echo "SPDX-License-Identifier: $license" is not a SPDX statement.

well that yes, but there are maybe variations which are not obviously to catch, because folks use SPDX-lic-ident-statements with invalid or non-captured license identifiers. So writing an exception for "if an SPDX lic identifiers starts with a special character ignores it" is maybe not as easy as having the escape blocks.

I would vote for starting simple and then different tags for sub cases could be added later. ... because maybe there are not many OSS projects which would use the feature itself and from these how many would like to express invalid copyright vs. unintended copyright statements?

carmenbianca:
Hi @pombredanne, @mcjaeger ! Thanks for the lengthy input. The reason we're looking at this is explained in #463, but I'll repeat here with some slight additional context.

We're given this slightly altered excerpt from [a real file in reuse-tool](https://github.com/fsfe/reuse-tool/blob/7ac2cd03b18a45463ae353763c7ef43757f510b6/src/reuse/_util.py#L48):

```python
_IDENTIFIER_PATTERN = re.compile(
    r"SPDX-License-Identifier: (.*?)" + _END_PATTERN, re.MULTILINE
)
_COPYRIGHT_PATTERNS = [
    re.compile(r"(SPDX-FileCopyrightText: .*?)" + _END_PATTERN),
    re.compile(r"(Copyright .*?)" + _END_PATTERN),
    re.compile(r"(© .*?)" + _END_PATTERN),
]
```

In a sort of meta way, all the regexes above find _themselves_ when run on the file. So you'll get some match like `.*?)" + _END_PATTERN),` as a copyright line, which is obviously ridiculous, and should be filtered out as a false positive.

Problem is that reuse-tool doesn't really do heuristics, and we don't currently do any smart code parsing. We don't know what is a comment and what isn't. It would be nice if we did (it would solve 99% of use cases for this problem), but we don't, and we almost certainly won't be able to support every single programming language.

Instead of making the tool smarter, we're proposing to give the dumb tool a dumb context clue that they should ignore the above code when it's wrapped in between `REUSE_IGNORE_BEGIN` and `REUSE_IGNORE_END` (or whatever we end up with).

Potential use cases for ignore blocks:

- As in #463, scripts that insert SPDX tags into a file.
- Documentation that describes the use of SPDX tags with examples. It's been a while since I last checked, but I believe this is the sole reason that the Linux kernel is not strictly REUSE compliant.
- A lot of internal plumbing in projects like reuse-tool that deal with SPDX tags. This tool's code base is a bit messy here and there to circumvent false positives.
- (Maybe?) Templates that contain SPDX tags.
- (Maybe?) Sanitation to prevent SPDX injection attacks ;)

---

Having said that, I'm not 100% sure whether those use cases are shared by SPDX generally.

mxmehl:
Thanks for chiming in Carmen!

> Having said that, I'm not 100% sure whether those use cases are shared by SPDX generally.

If more downstream users of SPDX would feel the same need that we do, I think chances are that we can bring it in (perhaps SPDX 3.0).

Anyway, for REUSE this issue is so pressing that we probably will introduce a REUSE specific tag anyway. I would be happy to deprecate/alias it with an SPDX tag later though.

mcjaeger:
@carmenbianca please help, from I wrote and what you wrote I have problems to understand, is that there a disagreement?

> we're proposing to give the dumb tool a dumb context clue that they should ignore the above code when it's wrapped in between REUSE_IGNORE_BEGIN and REUSE_IGNORE_END (or whatever we end up with).

In summary my post was about I am agreeing with this

carmenbianca:
@mcjaeger Ah, I was missing some of the context from the previous replies in this thread (adding hyper-specific ignore tags). Having re-read your comment now, I see we're in agreement :)

seabass-labrax:
@mxmehl, I noticed that you said that waiting for SPDX 3.0 wouldn't be ideal for REUSE. This month, we decided to release another minor version of SPDX, so we could maybe support this use-case in that if they don't need breaking changes :) 

Shall I ask to table this for a near-future SPDX Tech Team meeting?

mxmehl:
Oh sure, it might make sense to discuss that! I cannot promise that I can participate as Thursdays evenings are usually blocked, but I could make an exception if it's on the agenda.

seabass-labrax:
Great to hear that! Note that SPDX Tech Team meetings are on Tuesdays (18:00 CET), rather than the Legal Team's Thursdays.

pombredanne:
@carmenbianca re: https://github.com/fsfe/reuse-tool/pull/464#issuecomment-1035514571

Thank you for the background information! I still stand by my recommendation to stick to simpler way to ignore whole files or directories and not chunks within a file

floriansnow:
I think we shouldn't ignore whole files, but parts because that's the feature we need.  Ignoring anything is already not typically a good idea and since we need it, we should limit that as much as possible.

In regards to the naming of the tag:  If and when SPDX introduces such a tag, we can change ours to match theirs.  Until then, ours is something only REUSE uses and the name should not suggest more.  I also am not in favor of any of the suggestions so far because they imply other things than what we do.  We simply ignore what is in between those tags, so `REUSE_IGNORE_BEGIN` and `REUSE_IGNORE_END` are exactly what we need.  I'm not against other suggestions, just not the current ones because we don't do any escaping or something alike, so we shouldn't call it that.

floriansnow:
I think this is one of those cases where a vote is worse than building consensus because at least one of us is not ok with what some of those wordings would communicate.  I also agree with Carmen about going for consistency and in that regard, perhaps CamelCase should be consistent throughout as well, so `Reuse-IgnoreStart`.  But I'm ok with `Begin` or `Start` and in any capitalization.

mxmehl:
> I think this is one of those cases where a vote is worse than building consensus because at least one of us is not ok with what some of those wordings would communicate. 

Take it as a quick snapshot of the current opinions then. It helps to throw out options that no one would agree with anyway and focus on reaching a consensus between the two or three remaining opinions.

Regarding "Start" or "Begin" I have no strong opinion. If you think that Start is more common, happy to go with that then.

floriansnow:
@linozen Yes, my main concern was also just having something with `Ignore`.  So this is fine. :)

carmenbianca:
I'm good with that proposal @linozen

carmenbianca:
Can either of you implement the change?

linozen:
> Can either of you implement the change?

I can do it this afternoon :ok_person:

linozen:
This is ready to merge IMHO :tada:

src/reuse/_util.py
```diff
@@ -33,6 +35,9 @@
 GIT_EXE = shutil.which("git")
 HG_EXE = shutil.which("hg")
 
+REUSE_IGNORE_BEGIN = "REUSE-IgnoreStart"
+REUSE_IGNORE_END = "REUSE-IgnoreEnd"
```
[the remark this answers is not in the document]
linozen: true that. Will fix it now.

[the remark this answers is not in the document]
mxmehl: Thanks! There's still one on line 214 in _util.py. Just a comment, but still.

floriansnow (approved):

src/reuse/_util.py
```diff
@@ -204,6 +210,30 @@ def extract_spdx_info(text: str) -> SpdxInfo:
     return SpdxInfo(expressions, copyright_matches)
 
 
+def filter_ignore_block(text: str) -> str:
+    """Filter out blocks beginning with REUSE_IGNORE_BEGIN and ending with
```
mxmehl, line 214:
```suggestion
    """Filter out blocks beginning with REUSE_IGNORE_START and ending with
```

mxmehl (changes requested):
Otherwise LGTM

mxmehl:
Yes, ready to merge, thank you! Do you want to clean up the commits or shall we just merge?

linozen:
Cleaned up a little bit and will merge once the CI is happy.

linozen:
:tada:

PR #469, opened 2022-01-22
Add --recursive flag to addheader

### LINKED ISSUE #158: Feature: make `reuse addheader` able to work on multiple files/directories (--recursive?)
In my (ruby on rails) code, I know that all code under  `app/` is mine and can be licensed and copyrighted accordingly. Ideally instead of writing bash loops or using a similar approach, `reuse addheader` would implement a `--recursive` option to add headers to all files under a given path, or allow for globbing (`reuse addheader "app/*.rb"` or even `reuse addheader "app/**/*rb"`).

carmenbianca:
Fixes #158

### DIFF
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -615,13 +615,15 @@ def run(args, project: Project, out=sys.stdout) -> int:
 
     if args.recursive:
         paths = set()
-        all_files = project.all_files()
+        all_files = [path.resolve() for path in project.all_files()]
         for path in args.path:
             if path.is_file():
                 paths.add(path)
             else:
                 paths |= {
-                    child for child in all_files if path in child.parents
+                    child
+                    for child in all_files
+                    if path.resolve() in child.parents
                 }
     else:
         paths = args.path
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -358,19 +358,27 @@ def _verify_paths_line_handling(
 
 
 def _verify_paths_comment_style(paths: List[Path], parser: ArgumentParser):
+    unrecognised_files = []
+
     for path in paths:
         style = _get_comment_style(path)
         not_uncommentable = not _is_uncommentable(path)
 
         # TODO: This check is duplicated.
         if style is None and not_uncommentable:
-            parser.error(
+            unrecognised_files.append(path)
+
+    if unrecognised_files:
+        parser.error(
+            "{}\n{}".format(
                 _(
-                    "'{path}' does not have a recognised file extension,"
-                    " please use --style, --force-dot-license or"
-                    " --skip-unrecognised"
-                ).format(path=path)
+                    "The following files do not have a recognised file"
+                    " extension. Please use --style, --force-dot-license or"
+                    " --skip-unrecognised:"
+                ),
+                "\n".join(str(path) for path in unrecognised_files),
             )
+        )
 
 
 def _find_template(project: Project, name: str) -> Template:
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -575,7 +575,7 @@ def add_arguments(parser) -> None:
 
 def run(args, project: Project, out=sys.stdout) -> int:
     """Add headers to files."""
-    # pylint: disable=too-many-branches,too-many-locals
+    # pylint: disable=too-many-branches,too-many-locals,too-many-statements
     if not any((args.copyright, args.license)):
         args.parser.error(_("option --copyright or --license is required"))
 
@@ -606,16 +606,19 @@ def run(args, project: Project, out=sys.stdout) -> int:
         args.force_dot_license = True
 
     if args.recursive:
-        paths = []
+        paths = set()
         all_files = project.all_files()
         for path in args.path:
-            paths = paths + [
-                child for child in all_files if path in child.parents
-            ]
+            if path.is_file():
+                paths.add(path)
+            else:
+                paths |= {
+                    child for child in all_files if path in child.parents
+                }
     else:
         paths = args.path
 
-    paths = [_determine_license_path(path) for path in paths]
+    paths = {_determine_license_path(path) for path in paths}
 
     if not args.force_dot_license:
         _verify_write_access(paths, args.parser)
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -553,6 +553,14 @@ def add_arguments(parser) -> None:
         help=_("write a .license file instead of a header inside the file"),
     )
     parser.add_argument(
+        "--recursive",
+        "-r",
+        action="store_true",
+        help=_(
+            "add headers to all files under specified directories recursively"
+        ),
+    )
+    parser.add_argument(
         "--skip-unrecognised",
         action="store_true",
         help=_("skip files with unrecognised comment styles"),
@@ -567,7 +575,7 @@ def add_arguments(parser) -> None:
 
 def run(args, project: Project, out=sys.stdout) -> int:
     """Add headers to files."""
-    # pylint: disable=too-many-branches
+    # pylint: disable=too-many-branches,too-many-locals
     if not any((args.copyright, args.license)):
         args.parser.error(_("option --copyright or --license is required"))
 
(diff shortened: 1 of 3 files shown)


carmenbianca:
Consider automatically not failing (--skip-unrecognised) when doing recursive stuff

carmenbianca:
Consider adding --interactive

carmenbianca:
@floriansnow and I discussed --skip-unrecognised a little more, and we do not believe that this extra implied behaviour is desireable. If `reuse addheader ...`  does something, then `reuse addheader --recursive ...` should do exactly the same thing, but recursively. Adding more behaviour on top of that isn't amazing.

We want to explore other ways of making `reuse addheader --recursive --but-please-dont-break-on-me` a little easier.

carmenbianca:
Referring to #475 and #474 as ways of making --recursive nicer.

mxmehl:
OK, trying to understand the workflow you have been thinking of. Let's assume, a user runs the following command: `reuse addheader --recursive ./`:

1. If all files are recognised, headers will be added to all files
2. If only one file is not recognised, the whole process is being stopped, and not file is altered at all. The user is confronted with a helpful message that suggests the following:
    1. Re-run with `--interactive` to resolve the unrecognised files manually (style or .license)

As I user, I would also expect a `--skip-unrecognised` but as I read you don't think this is useful. Why? Many commands that allow batch operations also include an "ignore" flag.

Or did your discussion rather cover the question whether `--skip-unrecognised` should be the default behaviour of `--recursive`? In this case, I agree that is should not be automatically enabled, but hinted when problems appear (so case 2.).

carmenbianca:
> Or did your discussion rather cover the question whether --skip-unrecognised should be the default behaviour of --recursive? In this case, I agree that is should not be automatically enabled, but hinted when problems appear (so case 2.).

This is the correct interpretation. During the hackathon the group came to the opposite conclusion (i.e., `--skip-unrecognised` should be implied when `--recursive` is given). I did _not_ end up doing that in this PR.

`--skip-unrecognised` already exists, so `reuse addheader --recursive --skip-unrecognised ./` adds headers to all files except the single one from your example.

Improving output is a good suggestion. This is a recurring suggestion :)

mxmehl:
OK, so to finalise *this* PR, all it takes is a helpful output message mentioning `--skip-unrecognised`, right?

Later, this could be extended by `--interactive` once this function exists.

carmenbianca:
> OK, so to finalise _this_ PR, all it takes is a helpful output message mentioning `--skip-unrecognised`, right?

Well whaddaya know:

```
$ reuse addheader --copyright me foo.bar 
reuse.project - WARNING - could not find supported VCS
usage: reuse addheader [-h] [--copyright COPYRIGHT] [--license LICENSE] [--year YEAR]
                       [--style {applescript,aspx,bat,bibtex,c,css,f,ftl,handlebars,haskell,html,jinja,jsx,lisp,m4,ml,plantuml,python,rst,tex}]
                       [--copyright-style {spdx,spdx-symbol,string,string-c,string-symbol,symbol}] [--template TEMPLATE] [--exclude-year] [--single-line] [--multi-line]
                       [--explicit-license] [--skip-unrecognised] [--recursive]
                       path [path ...]
reuse addheader: error: 'foo.bar' does not have a recognised file extension, please use --style, --explicit-license or --skip-unrecognised
```

It already says that!

mxmehl:
> It already says that!

True, but only for the first file! I think the tool should collect all files that are not recognised and provide them at once. Otherwise it appears to the user that only this file is problematic. Perhaps I can extend the logic a bit (and you can polish it later ;)).

mxmehl:
Just did so, have fun ripping it apart ;)

```
$ ls foobar/
bar.nono  baz.sh  foo.py  foo.unknown  test.json

$ reuse addheader --recursive -c me -l MIT foobar/
usage: reuse addheader [-h] [--copyright COPYRIGHT] [--license LICENSE] [--year YEAR]
                       [--style {applescript,aspx,bat,bibtex,c,css,f,ftl,handlebars,haskell,html,jinja,jsx,lisp,m4,ml,plantuml,python,rst,tex}]
                       [--copyright-style {spdx,spdx-symbol,string,string-c,string-symbol,symbol}]
                       [--template TEMPLATE] [--exclude-year] [--single-line] [--multi-line]
                       [--explicit-license] [--skip-unrecognised] [--recursive]
                       path [path ...]
reuse addheader: error: The following files do not have a recognised file extension. No file has been edited. Please use --style, --explicit-license or --skip-unrecognised:
foobar/foo.unknown
foobar/bar.nono
```


To avoid only slightly different error messages, I also list the unrecognised files separately if only one file was targeted.

tests/test_main_addheader.py
```diff
@@ -939,3 +939,59 @@ def test_addheader_line_endings(
...
+    (fake_repository / "src/one/two").mkdir(parents=True)
+    (fake_repository / "src/one/two/foo.py").write_text(
+        cleandoc(
+            """
+            # spdx-License-Identifier: GPL-3.0-or-later
+            """
+        ).replace("spdx", "SPDX")
+    )
+    (fake_repository / "bar").mkdir(parents=True)
+    (fake_repository / "bar/bar.py").touch()
```
mxmehl, line 955: Just a note on what I detected during testing: reuse addheader (at least recursively) does not have an effect on empty files. It only picked them up when some content was in it. So this test may not catch all that you want (also making sure that nothing is written into `bar/bar.py`)

mxmehl:
I wanted to write a test for it, but failed:

```py
def test_addheader_recursive_contains_unrecognised(
    fake_repository, stringio, mock_date_today
):
    """Expect error and no edited files if at least one file has not been recognised."""
    (fake_repository / "baz").mkdir(parents=True)
    (fake_repository / "baz/foo.py").write_text("foo")
    (fake_repository / "baz/bar.unknown").write_text("bar")
    (fake_repository / "baz/baz.sh").write_text("baz")

    with pytest.raises(SystemExit):
        main(
            [
                "addheader",
                "--license",
                "Apache-2.0",
                "--copyright",
                "Jane Doe",
                "--recursive",
                "baz/",
            ]
        )

    assert "Jane Doe" not in (fake_repository / "baz/foo.py").read_text()
```

Error: `FAILED tests/test_main_addheader.py::test_addheader_recursive_contains_unrecognised - Failed: DID NOT RAISE <class 'SystemExit'>`

Running this command certainly exits with a non-zero code just like running the same command over a single unrecognised file. Why does pylint not detect the error?

src/reuse/header.py
```diff
@@ -351,19 +351,26 @@ def _verify_paths_line_handling(
...
-                _(
-                    "'{path}' does not have a recognised file extension,"
-                    " please use --style, --explicit-license or"
-                    " --skip-unrecognised"
-                ).format(path=path)
-            )
+            unrecognised_files.append(path)
+
+    # If at least one file was unrecognised, print error
+    if len(unrecognised_files) > 0:
```
carmenbianca, line 366: Lists in Python are truthy :) The conventional way to do this is `if unrecognised_files`

src/reuse/header.py
```diff
@@ -351,19 +351,26 @@ def _verify_paths_line_handling(
...
+        parser.error(
+            _(
+                "The following files do not have a recognised file extension."
+                " No file has been edited. Please use --style,"
+                " --explicit-license or --skip-unrecognised:\n{paths}"
+            ).format(paths="\n".join([str(i) for i in unrecognised_files]))
```
carmenbianca, line 372: `i` is reserved for integers. `path` would be a better temporary value here.

src/reuse/header.py
```diff
@@ -351,19 +351,26 @@ def _verify_paths_line_handling(
...
-            )
+            unrecognised_files.append(path)
+
+    # If at least one file was unrecognised, print error
+    if len(unrecognised_files) > 0:
+        parser.error(
+            _(
+                "The following files do not have a recognised file extension."
+                " No file has been edited. Please use --style,"
+                " --explicit-license or --skip-unrecognised:\n{paths}"
```
carmenbianca, line 371: I'm not entirely happy with `\n{paths}` in the translatable string. If it weren't a translatable string, it'd be fine, but it's a good idea to not put formatting in translatable strings if you can help it.

src/reuse/header.py
```diff
@@ -351,19 +351,26 @@ def _verify_paths_line_handling(
...
-            parser.error(
-                _(
-                    "'{path}' does not have a recognised file extension,"
-                    " please use --style, --explicit-license or"
-                    " --skip-unrecognised"
-                ).format(path=path)
-            )
+            unrecognised_files.append(path)
+
+    # If at least one file was unrecognised, print error
```
carmenbianca, line 365: Misleading comment. It doesn't just print an error; it exits.

carmenbianca (commented):
Review! I'll adjust the suggestions myself.

mxmehl:
Thanks for the suggestions, all make absolutely sense.

AFAICS all left to do would be the tests.

mxmehl (approved):

PR #476, opened 2022-01-22
Rename --explicit-license to --force-dot-license and improve its help text

### LINKED ISSUE #473: Improve help string of addheader --explicit-license

floriansnow:
Closes #473.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,13 @@ The versions follow [semantic versioning](https://semver.org).
 - Use `setuptools` instead of the deprecated `distutils` which will be removed
   with Python 3.12 (#451)
 
+- `addheader --explicit-license` renamed to `--force-dot-license`. (#476)
+
 ### Deprecated
+
+- Deprecated `--explicit-license` in favour of `--force-dot-license`.
+  `--explicit-license` will remain useable (although undocumented) for the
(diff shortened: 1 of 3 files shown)


floriansnow:
Looking at this option, I feel like `--explicit-license` is not a good fit either.  I didn't change it because it's a breaking change and I'm not sure about the process here, but please let me know if I should just add it here or create a separate issue for something like that.

carmenbianca:
If you know a better name, we can change it such that the old name will still work, but only the new name will be presented to the user. Then we just make mention of it in the CHANGELOG

floriansnow:
My suggestions are `--external-license` or `--license-file` or `--external-license-file` in no particular order of preference.

carmenbianca:
What about `--force-dot-license`?

floriansnow:
Sure, I like that too.

carmenbianca:
@floriansnow How do the committed changes look?

floriansnow:
Looks very good to me!  Thank you!

mxmehl:
LGTM, but why not mentioning the deprecated command in the "Deprecated" section of the changelog? ;)

carmenbianca (approved):

mxmehl (approved):

PR #478, opened 2022-01-22
Tidy up workflows and extend by multi-platform Docker image push

### LINKED ISSUE #449: Provide an aarch64 container with reuse
The `fsfe/reuse` container is only available for amd64. It would be useful to have an aarch64 build as well. 

https://hub.docker.com/r/fsfe/reuse/tags

carmenbianca:
- Move Docker out of pythonpackage
- Better describe latesttag
- Make docker workflow also push the Docker images to Dockerhub (as they made autobuild cost money)
- Publish docker images as amd64 and arm64. Fixes #449

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -13,7 +13,7 @@ on:
     # - 1 (not with 0 though)
     # - latest"
     tags:
-      - 'v*.*.*'
+      - "v*.*.*"
     # Main branch will be "dev"
     branches:
       - main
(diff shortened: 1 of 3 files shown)


mxmehl:
I would like to combine that with the new workflows for pushing the Docker images for which I wanted to streamline the whole layout anyway. So thanks for starting this!

carmenbianca:
@mxmehl OK, feel free to push to this branch whenever!

mxmehl:
I think it's done. Tested this on a separate repo extensively to make it do what we want (see the updated first post).

There are two TODOs for when we renamed the master branch to main. Also, I'll have to check whether we can run this action for the latest tag to push the new arm64 images, or whether this has to wait for the next version release.

Also, I am not sure whether we need the latesttag check any more since the process is now automated. Actually, we can also save ourselves the hassle with the manual git re-tagging of "latest".

PR #479, opened 2022-01-23
Rename `master` to `main` and communicate the change :unicorn: 

### LINKED ISSUE #459: Rename master branch to main
More and more projects are making this change, and it's probably a good new default.

I suspect that this has some knock-on effects here and there. Good first issue, not very high priority.

Try this in a repo of a user/group that _isn't_ @fsfe first.

Arghya-Banerjee:
I have tried to create and rename the default branch as "main" as you requested. Please look at it if it works. Thank You

Fixes #459

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ docker run --rm --volume $(pwd):/data fsfe/reuse --include-submodules spdx -o ou
 There are a number of tags available:
 
 - `latest` is the most recent stable release.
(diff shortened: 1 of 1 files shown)


carmenbianca:
Thank you for your contribution @Arghya-Banerjee !

I think this change is best done after the v0.15.0 milestone. So this will probably stay in the fridge for a little while :)

floriansnow:
@carmenbianca What's your concern if we were to do this now?

PR #480, opened 2022-01-23
Create --skip-existing flag for addheader

### LINKED ISSUE #474: Create reuse addheader --skip-existing
Skip files that already have header information

carmenbianca:
Fixes #474

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@ The versions follow [semantic versioning](https://semver.org).
 
 - Recommendations for installation/run methods: package managers and pipx (#457)
 
+- Added `--skip-existing` flag to `addheader` in order to skip files that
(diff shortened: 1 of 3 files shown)


mxmehl:
LGTM. But shall we give some (optional) info message on that this file has been skipped?

mxmehl:
> I think the info message might make sense. But I wouldn't make it optional because that would mean yet another flag, unless we included it in a more general `quiet` flag.

Full ack. I think we already have a --quiet flag for lint. In general, we could think about introducing both quiet and verbose flags for some commands that help debugging or turning down noise for large repos and operations.

For an intermediate solution, what about adding this info message, and dealing with verboseness later?

mxmehl:
Added this info message, very similar to the one from --skip-unrecognised

For me, that would be good to go.

carmenbianca:
Merged manually.

PR #484, opened 2022-01-23
WIP: Bump versions in requirements[-dev].txt

carmenbianca:


mxmehl:
Oh yes, I remember that bumping pylint caused some errors in an earlier dependency bump I did.

mxmehl:
In the Github workflows we are also running into https://github.com/pallets/markupsafe/issues/284

Not sure whether this PR solves it already, but we should think about pinning all dependencies to avoid such errors.

carmenbianca:
Closed by #491

PR #485, opened 2022-01-24
Extend .prettierignore

mxmehl:
Some people have a `/venv` which makes `make prettier` go though all files there.

You could theoretically also make prettier just be based on `.gitignore`, but that would then not ignore the SPDX files.

So given this, the while .prettierignore file probably has to be extended to catch locally ignored files, e.g. .Python, dist, lib, .cache or whatever, that are currently caught by .gitignore.

@carmenbianca What do you think?

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,4 +2,19 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+# SPDX resources
 /src/reuse/resources/*.json
+
+# Additional paths in which third-party code/files are located
+.env
+.pytest_cache/
+.tox/
+.venv
+build/
+dist/
+docs/_build/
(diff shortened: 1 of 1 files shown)


carmenbianca:
I tried `prettier --ignore-path .prettier-ignore --ignore-path --check .gitignore .`, but it only uses the last specified ignore path.

Using `.gitignore` is not an option as you observed.

I also noticed that `.tox` is a conflicting directory. `build/`, `dist/`, and `docs/_build` are also potentially painful. Furthermore, `.gitignore` specifies these paths:

```gitignore
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/
```

I think it's fine to manually add these to `.prettierignore`.

PR #489, opened 2022-01-26
Make license_list_up_to_date.py actually fail when it fails

### LINKED ISSUE #487: Should license_list_up_to_date.yaml action fail?
We have a daily job checking whether the SPDX license lists are up-to-date. The job seems to be successful even when they aren't, so one would have to look in the logs.

Shall it rather fail if on of them is outdated?

carmenbianca:
Fixes #487

### DIFF
--- a/.github/workflows/latesttag.py
+++ b/.github/workflows/latesttag.py
@@ -4,6 +4,7 @@
 
 import json
 import subprocess
+import sys
 from distutils.version import LooseVersion
 from urllib.request import urlopen
 
@@ -31,4 +32,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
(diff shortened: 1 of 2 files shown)


carmenbianca:
Add change log entry.

mxmehl (approved):

carmenbianca:
@mxmehl I changed this PR quite a bit. It pulled spdx/license-list-data's master branch, which was not ideal. It now pulls the latest version tag instead.

I'll merge this when the tests pass.

mxmehl:
Wait, but then it does not match what we are doing in the Makefile where we pull from master, doesn't it?

https://github.com/fsfe/reuse-tool/blob/7c3cc6eb8acd7676e13b3bdda40c5bfc1a8fca82/Makefile#L128-L133

PR #491, opened 2022-01-28
Improve pylint config

floriansnow:
This should not change any functionality, just improve code quality and enable better future code quality by using more sensible checks.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,55 +1,124 @@
 # SPDX-FileCopyrightText: 2018 Free Software Foundation Europe e.V. <https://fsfe.org>
+# SPDX-FileCopyrightText: 2022 Florian Snow <florian@familysnow.net>
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 [MASTER]
-
 jobs=0
+
 
 [MESSAGES CONTROL]
 
-# Disable the message, report, category or checker with the given id(s). You
-# can either give multiple identifiers separated by comma (,) or put this
-# option multiple times (only on the command line, not in the configuration
-# file where it should appear only once).You can also use "--disable=all" to
-# disable everything first and then reenable specific checks. For example, if
-# you want to run only the similarities checker, you can use "--disable=all
-# --enable=similarities". If you want to run only the classes checker, but have
-# no Warning level messages displayed, use"--disable=all --enable=classes
-# --disable=W"
-disable=redefined-builtin,C0330,duplicate-code,logging-format-interpolation,line-too-long,implicit-str-concat,logging-fstring-interpolation,unsubscriptable-object,inherit-non-class
+disable=R0801,
+         logging-fstring-interpolation,
+enable=useless-suppression
 
 [REPORTS]
 
-# Set the output format. Available formats are text, parseable, colorized, json
-# and msvs (visual studio).You can also give a reporter class, eg
-# mypackage.mymodule.MyReporterClass.
 output-format=text
-
-# Tells whether to display a full report or only the messages
 reports=yes
+
 
 [FORMAT]
 
-# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
 expected-line-ending-format=LF
 
-# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
-# tab).
+ignore-long-lines=^\s*(# )?<?https?://\S+>?$|^# SPDX-.*$
+indent-after-paren=4
 indent-string='    '
-
-# Maximum number of characters on a single line.
-max-line-length=79
-
-# Allow the body of an if to be on the same line as the test if there is no
-# else.
+max-line-length=80
 single-line-if-stmt=no
+
+
+[REFACTORING]
+
+max-nested-blocks=4
+never-returning-functions=sys.exit
+
 
 [BASIC]
 
-# Good variable names which should always be accepted, separated by a comma
+argument-naming-style=snake_case
+attr-naming-style=snake_case
+bad-names=foo,
+          bar,
+          baz,
+          toto,
+          tutu,
+          tata
+class-attribute-rgx=^[a-z0-9_]*$|^[A-Z0-9_]*$
+class-naming-style=PascalCase
+const-naming-style=UPPER_CASE
+docstring-min-length=3
+function-naming-style=snake_case
 good-names=i,j,k,ex,Run,_,fp
+inlinevar-naming-style=snake_case
+method-naming-style=snake_case
+module-naming-style=snake_case
+variable-naming-style=snake_case
+
 
 [MISCELLANEOUS]
 
 notes=FIXME,XXX
+
+
+[SIMILARITIES]
+
+ignore-comments=yes
+ignore-docstrings=yes
+ignore-imports=yes
+min-similarity-lines=4
+
+
+[SPELLING]
+
+max-spelling-suggestions=4
+spelling-store-unknown-words=no
+
+
+[STRING]
+
+check-quote-consistency=yes
+check-str-concat-over-line-jumps=yes
+
+
+[DESIGN]
+
+max-args=6
+max-attributes=10
+max-bool-expr=5
+# we probably want to reduce this a bit, after some refactoring
+max-branches=12
+max-locals=15
+max-parents=7
+max-public-methods=20
+max-returns=6
+max-statements=50
+min-public-methods=1
+
+
+[IMPORTS]
+allow-wildcard-with-all=no
+analyse-fallback-blocks=no
+deprecated-modules=optparse,tkinter.tix
+known-third-party=enchant
+
+
+[CLASSES]
+defining-attr-methods=__init__,
+                      __new__,
+                      setUp,
+                      __post_init__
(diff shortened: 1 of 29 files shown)


src/reuse/_util.py
```diff
@@ -222,8 +225,8 @@ def make_copyright_line(
     copyright_prefix = _COPYRIGHT_STYLES.get(copyright_style)
     if copyright_prefix is None:
         raise RuntimeError(
-            "Unexpected copyright style: Need 'spdx', 'spdx-symbol', 'string', 'string-c',"
-            "'string-symbol' or 'symbol'"
+            "Unexpected copyright style: Need 'spdx', 'spdx-symbol', 'string',"
+            "'string-c', 'string-symbol' or 'symbol'"
```
carmenbianca, line 228: missing space between 'string' and 'string-c': `'string,'string-c', 'string-symbol'`

.pylintrc
```diff
@@ -1,55 +1,124 @@
...
 
 [REPORTS]
 
-# Set the output format. Available formats are text, parseable, colorized, json
-# and msvs (visual studio).You can also give a reporter class, eg
-# mypackage.mymodule.MyReporterClass.
 output-format=text
-
-# Tells whether to display a full report or only the messages
 reports=yes
```
carmenbianca, line 19: Maybe this line should be removed to get a less verbose output.

.pylintrc
```diff
@@ -1,55 +1,124 @@
...
 
-# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
 expected-line-ending-format=LF
 
-# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
-# tab).
+ignore-long-lines=^\s*(# )?<?https?://\S+>?$|^# SPDX-.*$
+indent-after-paren=4
 indent-string='    '
+max-line-length=80
```
carmenbianca, line 29: I believe this is set to 79 in pyproject.toml.

.pylintrc
```diff
@@ -1,55 +1,124 @@
...
 
-# Maximum number of characters on a single line.
-max-line-length=79
 
-# Allow the body of an if to be on the same line as the test if there is no
-# else.
-single-line-if-stmt=no
+[REFACTORING]
+
+max-nested-blocks=4
```
carmenbianca, line 35: I'm confused on this one. Why is pylint succeeding when we definitely have some deeply nested blocks?

.pylintrc
```diff
@@ -1,55 +1,124 @@
...
+max-attributes=10
+max-bool-expr=5
+# we probably want to reduce this a bit, after some refactoring
+max-branches=12
+max-locals=15
+max-parents=7
+max-public-methods=20
+max-returns=6
+max-statements=50
+min-public-methods=1
```
carmenbianca, line 98: I'm not sure how useful this all is. I know pylint has some defaults for these, and they're a bit stricter than these, but I've sort of resigned to the idea that some bits of code are just going to go over the limits, and that's okay.

Maybe there's something I need to read somewhere, but I cannot for the life of me write tidy, small main functions.

.pylintrc
```diff
@@ -1,55 +1,124 @@
...
+
+
+[IMPORTS]
+allow-wildcard-with-all=no
+analyse-fallback-blocks=no
+deprecated-modules=optparse,tkinter.tix
```
carmenbianca, line 104: distutils

src/reuse/_util.py
floriansnow: Good point!  I went with the same style that was there before and replicated the missing space.  :facepalm:  Fixed in 425904b08f74b16dc6b4e93c503e0cdda777df35.

.pylintrc
floriansnow: Yeah, I prefer that too.  Didn't want to overrule your previous choice, though.  Fixed in 205fd7d73d9e16a9beb2d5c478fc00a4484c9565.

.pylintrc
floriansnow: Good catch!  Fixed in 5f2e9d67cdfa27ff3c460fc6f0c3f7e79ff14927.

.pylintrc
floriansnow: Some have this check explicitly disabled.  I removed a lot of local disables, but none that would have required larger refactorings.  So I guess that's the answer.  Or do you have some others in mind?

.pylintrc
floriansnow: Most of these are the defaults and I just wanted to make them more explicit (otherwise, you'd always have to check the documentation).

In regards to some code violating these rules: Yes, that can be fine.  That what local ignores are for.  It's still good to have the checks because they do point to potential code rot.  In some cases, there may be good reasons for breaking those rules, especially because they're rules of thumb anyway, but it's a good idea to pause and think again when pylint complains.

Small main functions can indeed be tricky and I struggle with it sometimes too.  But it is a matter of design and I am getting better at it, so I think as long as we move in that direction, it's fine.  :-)

.pylintrc
floriansnow: Added in 6bf2d087a552841ee319d6302145c707bde87b8e.

.pylintrc
carmenbianca: Could you comment out the defaults? Or otherwise mark them as being default?

floriansnow: I could, but they may not be the defaults in future versions, so I think it's better to just have them as is.

.pylintrc
carmenbianca: Just tested locally; apparently there really isn't a lot of nesting in the code base :o Go team.

.pylintrc
carmenbianca: OK fair :)

carmenbianca:
During some local testing, I ran into this:

```
$ make lint
pylint src/reuse tests/*.py
************* Module reuse._licenses
src/reuse/_licenses.py:27:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
src/reuse/_licenses.py:42:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module reuse.lint
src/reuse/lint.py:339:14: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
src/reuse/lint.py:339:14: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
************* Module reuse.supported_licenses
src/reuse/supported_licenses.py:29:12: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module reuse.download
src/reuse/download.py:73:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module reuse.project
src/reuse/project.py:61:42: R1735: Consider using {} instead of dict() (use-dict-literal)
src/reuse/project.py:234:24: R1735: Consider using {} instead of dict() (use-dict-literal)
************* Module reuse.report
src/reuse/report.py:61:24: R1735: Consider using {} instead of dict() (use-dict-literal)
src/reuse/report.py:62:32: R1735: Consider using {} instead of dict() (use-dict-literal)
src/reuse/report.py:63:28: R1735: Consider using {} instead of dict() (use-dict-literal)
src/reuse/report.py:67:42: R1735: Consider using {} instead of dict() (use-dict-literal)
src/reuse/report.py:192:19: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
src/reuse/report.py:355:38: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module reuse._util
src/reuse/_util.py:39:15: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/reuse/_util.py:42:12: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/reuse/_util.py:344:22: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module test_main_addheader
tests/test_main_addheader.py:961:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module test_util
tests/test_util.py:44:21: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)

------------------------------------------------------------------
Your code has been rated at 9.94/10 (previous run: 9.94/10, +0.00)
```

This was run with `pylint==2.6.0`, because I hadn't upgraded yet.

The above warnings no longer appear. I'm not all too bothered about some of them, but I think that `unspecified-encoding` (#455) and `consider-using-with` are valuable.

Thing is, I can't get those warnings to show up at all:

```diff
diff --git a/.pylintrc b/.pylintrc
index 2f9bff9..daa1281 100644
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,7 +11,9 @@ jobs=0
 
 disable=R0801,
          logging-fstring-interpolation,
-enable=useless-suppression
+enable=useless-suppression,
+       unspecified-encoding,
+       consider-using-with,
 
 [REPORTS]
 ```

```
$ make lint
pylint src/reuse tests/*.py

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

What gives? :thinking:

carmenbianca:
Figured it out. It's a bug in `pylint~=2.7`. Any reason we're not using `pylint~=2.12`?

PR #494, opened 2022-01-31
Ignore more SPDX file formats

### LINKED ISSUE #442: Ignore all SPDX files in repo (not just the tag-value flavour)
Currently `reuse lint` ignores `*.spdx` extensions (i.e. tag-value SPDX format), which AFAICT is intentional.

I noticed though that it treats `*-spdx.rdf` (i.e. the RDF SPDX format) as a normal file and want it to have REUSE-compliant headers. I suspect the same will hold true for the JSON and YAML flavors as well.

As such, I suggest that `reuse lint` ignore all SPDX formats:
```
.spdx
.spdx.rdf
.spdx.json
.spdx.yaml
.spdx.yml
```

I suspect it falls under `_IGNORE_FILE_PATTERNS` in:
https://github.com/fsfe/reuse-tool/blob/4ce4c5fe2976536e9f2911048f1b6b308fa191ee/src/reuse/__init__.py#L48

See: https://spdx.github.io/spdx-spec/conformance/#44-standard-data-format-requirements

mxmehl:
Fixes #442

The regexes could also be combined to `.*\.spdx(.(rdf|json|xml|ya?ml))?$` but I figured that this may rather increase confusion. One could also simplify `ya?ml` to `yaml|yml` but I think that's still an easy one.

### DIFF
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -53,11 +53,20 @@ _IGNORE_FILE_PATTERNS = [
     re.compile(r"^\.gitkeep$"),
     re.compile(r"^\.hgtags$"),
     re.compile(r".*\.license$"),
-    re.compile(r".*\.spdx$"),
     # Workaround for https://github.com/fsfe/reuse-tool/issues/229
     re.compile(r"^CAL-1.0(-Combined-Work-Exception)?(\..+)?$"),
     re.compile(r"^SHL-2.1(\..+)?$"),
 ]
+
+_IGNORE_SPDX_PATTERNS = [
+    # SPDX files from
+    # https://spdx.github.io/spdx-spec/conformance/#44-standard-data-format-requirements
+    re.compile(r".*\.spdx$"),
+    re.compile(r".*\.spdx.(rdf|json|xml|ya?ml)$"),
+]
+
+# Combine SPDX patterns into file patterns to ease default ignore usage
+_IGNORE_FILE_PATTERNS.extend(_IGNORE_SPDX_PATTERNS)
 
 #: Simple structure for holding SPDX information.
 #:
--- a/src/reuse/spdx.py
+++ b/src/reuse/spdx.py
@@ -9,6 +9,7 @@ import logging
 import sys
 from gettext import gettext as _
 
+from . import _IGNORE_SPDX_PATTERNS
 from ._util import PathType
 from .project import Project
 from .report import ProjectReport
@@ -28,9 +29,19 @@ def run(args, project: Project, out=sys.stdout) -> int:
     with contextlib.ExitStack() as stack:
         if args.file:
             out = stack.enter_context(args.file.open("w", encoding="utf-8"))
-            if args.file.suffix != ".spdx":
+            spdxfile = [
+                pattern
+                for pattern in _IGNORE_SPDX_PATTERNS
+                if pattern.match(args.file.name)
+            ]
+            if not spdxfile:
+                # pylint: disable=line-too-long
                 _LOGGER.warning(
-                    _("'{path}' does not end with .spdx").format(path=out.name)
+                    _(
+                        "'{path}' does not match a common SPDX file pattern. Find"
+                        " the suggested naming conventions here:"
+                        " https://spdx.github.io/spdx-spec/conformance/#44-standard-data-format-requirements"
+                    ).format(path=out.name)
                 )
 
         report = ProjectReport.generate(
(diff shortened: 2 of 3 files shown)


carmenbianca (changes requested):
PR looks good. One thing is missing is changes to [this line](https://github.com/fsfe/reuse-tool/blob/10941da1c348e3bf7f4827c9f2efafa8b7236a83/src/reuse/spdx.py#L32).

mxmehl:
Thanks for the review, and good catch. Our spdx subcommand does not output the formats specified with SPDX 2.2, but it's good to catch them already.

I figured two ways to match them, each one in separate commits (the first one has a unhelpful log message, please ignore). The first one uses the same technique as before, but due to the different cases we have to check it's quite complex. The second approach is more elegant I think. It also has the advantage that we keep the file name patterns for ignored SPDX files in sync.

Please let me know what you prefer.

I didn't look into tests yet.

src/reuse/__init__.py
```diff
@@ -53,14 +53,17 @@
...
     # Workaround for https://github.com/fsfe/reuse-tool/issues/229
     re.compile(r"^CAL-1.0(-Combined-Work-Exception)?(\..+)?$"),
     re.compile(r"^SHL-2.1(\..+)?$"),
 ]
 
+_IGNORE_SPDX_PATTERNS = [
+    # SPDX files from https://spdx.github.io/spdx-spec/conformance/#44-standard-data-format-requirements
+    re.compile(r".*\.spdx$"),
+    re.compile(r".*\.spdx.(rdf|json|xml|ya?ml)$"),
+]
```
carmenbianca, line 65: I like this idea a lot. But I think we can reduce the likelihood of errors like this:

```python
_IGNORE_SPDX_PATTERN = re.compile(r"combined pattern goes here")

_IGNORE_FILE_PATTERNS = [
    re.compile(r"^LICENSE"),
    [...],
    _IGNORE_SPDX_PATTERN,
]
```

This way you don't have to specifically check against `_IGNORE_FILE_PATTERNS + _IGNORE_SPDX_PATTERNS` elsewhere.

src/reuse/spdx.py
```diff
@@ -27,17 +28,17 @@ def run(args, project: Project, out=sys.stdout) -> int:
...
+        spdxfile = [
+            pattern
+            for pattern in _IGNORE_SPDX_PATTERNS
+            if pattern.match(args.file.name)
+        ]
+        if not spdxfile:
+            _LOGGER.warning(
+                _("'{path}' does not match a known SPDX file pattern").format(
+                    path=out.name
+                )
```
carmenbianca, line 40: I think this message is too abstract. Maybe something like: `"'{path}' does not end with .spdx or .spdx.ext"`, but I'm not a fan of that either.

Maybe provide a URL describing SPDX document file extensions or something?

mxmehl: Sure, attempted it in another commit

floriansnow: LGTM!

src/reuse/__init__.py
floriansnow: @mxmehl @carmenbianca How can we move this along?  I think either version is fine.

carmenbianca: The proposal in @mxmehl 's last comment LGTM.

carmenbianca:
Don't forget change log entry.

mxmehl:
Rebased, implemented my last suggestion, and added a changelog. Please review again :)

src/reuse/spdx.py
```diff
@@ -28,9 +29,19 @@ def run(args, project: Project, out=sys.stdout) -> int:
     with contextlib.ExitStack() as stack:
         if args.file:
             out = stack.enter_context(args.file.open("w", encoding="utf-8"))
-            if args.file.suffix != ".spdx":
+            spdxfile = [
+                pattern
+                for pattern in _IGNORE_SPDX_PATTERNS
+                if pattern.match(args.file.name)
+            ]
+            if not spdxfile:
```
carmenbianca, line 37:
```suggestion
            if not any(
                pattern.match(args.file.name)
                for pattern in _IGNORE_SPDX_PATTERNS 
            ):
```

carmenbianca (dismissed):

src/reuse/spdx.py
mxmehl: Elegant! Implemented in baa1f2f27bb0d0b9b24676fa12d2209513b70567

floriansnow (approved):

PR #496, opened 2022-02-03
ignore meson subproject directories

ventosus:
Meson subprojects should be handled similarly to git submodules. Meson
subprojects may actually be git submodules, but they may also just be
regular directories e.g. included via git subtree or via other means.

If the project of question uses meson as its build system, e.g. has the
following file:

    ./meson.build

Then directories in the subprojects root directory should be ignored by
default:

    ./subprojects/libfoo/foo.c
    ./subprojects/libbar/bar.c

Normal files in the subprojects root directory are normally parsed:

    ./subprojects/libfoo.wrap
    ./subprojects/libbar.wrap

This can be overriden with the command line switch:

    reuse --include-meson-subprojects lint

Relevant resources are at [1] and [2].

[1] https://mesonbuild.com/Subprojects.html
[2] https://mesonbuild.com/Wrap-dependency-system-manual.html

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -94,6 +94,97 @@ def test_lint_submodule_included_fail(submodule_repository, stringio):
     assert ":-(" in stringio.getvalue()
 
 
+def test_lint_meson_subprojects(fake_repository, stringio):
+    """Run a successful lint."""
+    (fake_repository / "meson.build").write_text(
+        "SPDX-FileCopyrightText: 2022 Jane Doe\n"
+        "SPDX-License-Identifier: CC0-1.0"
+    )
+    subprojects_dir = fake_repository / "subprojects"
+    subprojects_dir.mkdir()
+    libfoo_dir = subprojects_dir / "libfoo"
+    libfoo_dir.mkdir()
+		# ./subprojects/foo.wrap has license and linter succeeds
+    (subprojects_dir / "foo.wrap").write_text(
+        "SPDX-FileCopyrightText: 2022 Jane Doe\n"
+        "SPDX-License-Identifier: CC0-1.0"
+    )
+    # ./subprojects/libfoo/foo.c misses license but is ignored
+    (libfoo_dir / "foo.c").write_text("foo")
+    result = main(["lint"], out=stringio)
+
+    assert result == 0
+    assert ":-)" in stringio.getvalue()
+
+
+def test_lint_meson_subprojects_fail(fake_repository, stringio):
+    """Run a successful lint."""
+    (fake_repository / "meson.build").write_text(
+        "SPDX-FileCopyrightText: 2022 Jane Doe\n"
+        "SPDX-License-Identifier: CC0-1.0"
+    )
+    subprojects_dir = fake_repository / "subprojects"
+    subprojects_dir.mkdir()
+    libfoo_dir = subprojects_dir / "libfoo"
+    libfoo_dir.mkdir()
+		# ./subprojects/foo.wrap misses license and linter fails
+    (subprojects_dir / "foo.wrap").write_text("foo")
+    # ./subprojects/libfoo/foo.c misses license but is ignored
+    (libfoo_dir / "foo.c").write_text("foo")
+    result = main(["lint"], out=stringio)
+
+    assert result == 1
+    assert ":-(" in stringio.getvalue()
+
+
+def test_lint_meson_subprojects_included_fail(fake_repository, stringio):
+    """Run a successful lint."""
+    (fake_repository / "meson.build").write_text(
+        "SPDX-FileCopyrightText: 2022 Jane Doe\n"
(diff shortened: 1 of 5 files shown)


mxmehl:
Thanks for your suggestion and the code!

Before looking at the code too closely (for which others are better suited anyway), I am not sure whether we want to have this flag enabled by default like the submodules. I assume there are other projects out there that contain a `subproject` directory without using meson at all.

I comparison, Git submodules are excluded by default by the REUSE spec, so it makes totally sense to also exclode them in the tool.

tests/test_main.py
```diff
@@ -93,6 +93,97 @@ def test_lint_submodule_included_fail(submodule_repository, stringio):
     assert ":-(" in stringio.getvalue()
 
 
+def test_lint_meson_subprojects(fake_repository, stringio):
+    """Run a successful lint."""
+    (fake_repository / "meson.build").write_text(
+        "SPDX-FileCopyrightText: 2022 Jane Doe\n"
```
mxmehl, line 99: This probably throws a number of lint errors. Please escape it in some way, e.g. by splitting the string or making it lowercase (and uppercase later). There are a number of examples in other tests.

Also applies to lines below, e.g. 108/109.

mxmehl (changes requested):

tests/test_main.py
ventosus: Sorry, I've no idea what the problem is, could you be more specific?

    pytest -k meson

runs through without errors...

ventosus:
> I assume there are other projects out there that contain a `subproject` directory without using meson at all.

The code of course checks whether meson is used as build system, and only then excludes directories in the *subprojects* folder.

> I comparison, Git submodules are excluded by default by the REUSE spec, so it makes totally sense to also exclode them in the tool.

Maybe I'm blind, but I cannot see any mention of git submodules in the spec to be excluded from linting Only files that are part of version control and specifically excluded from version control are skipeed while linting. git submodules *are* under version control.

tests/test_main.py
mxmehl: Yes, but running `reuse lint` over the project fails. This is also shown in the Actions (also `black` seems to demand changes).

You could escape it by this:
https://github.com/fsfe/reuse-tool/blob/7ac2cd03b18a45463ae353763c7ef43757f510b6/tests/test_project.py#L106-L108

mxmehl:
> The code of course checks whether meson is used as build system, and only then excludes directories in the _subprojects_ folder.

Ah OK, I see. @carmenbianca @silverhook what do you think about such a specific ignore?

> Maybe I'm blind, but I cannot see any mention of git submodules in the spec to be excluded from linting .Only files that are part of version control or specifically excluded from version control are skipeed while linting. git submodules _are_ under version control.

You are right, this issue is still open but the path is clear: the spec will ignore Git submodules. fsfe/reuse-docs#36

ventosus:
> You are right, this issue is still open but the path is clear: the spec will ignore Git submodules. fsfe/reuse-docs#36

I see. This solves the issue for those that include subprojects via git submodules but not for those that include subprojects via other means, e.g. git subtree or verbatim copy.

For a meson project it's pretty clear that every directory und ./subprojects *is* a subproject. If the *--recursive* flag will ever come around, it could make sense to subsum meson subprojects with git submodules.

Why have I come up with this patch in the first place?

Really a lot of my projects include subprojects via *git submodule* and  look like this, e.g. they're all esoteric in @carmenbianca's terms ;-)

*projA
 * meson.build
 * LICENSES
  * MIT 
 * src
  * foo.c 
 * subprojects
  * projA 
   * meson.build
   * LICENSES
    * MIT
    * ISC
   * .reuse
    * dep5 (Files ext / * ISC)
   * src
    * bar.c
   * ext
    * hi.c
As the subprojects are mine, too, the parent project would comply to REUSE, but in order to do so, I need to duplicate the dep5 files from all subprojects which is superfluous, as the information is already there.

Maybe there's a better solution for my use case than this patch here? I don't know.

ventosus:
As a side note, all my subprojects are checked for and pass REUSE compatibility by their own as part of the build/test mechanism, so no information is lost by excluding them from linting of the parent project.

tests/test_main.py
carmenbianca: Fixed in [78e172d](https://github.com/fsfe/reuse-tool/pull/496/commits/78e172d02674745edeb1212b6ff904ff3a8750de)

carmenbianca:
Pushed some changes/fixes to the branch; now LGTM.

Thanks @ventosus !

Ready for review.

carmenbianca:
Ah, this still needs a change log entry.

mxmehl (approved):
LGTM, but yes, please do the changelog entry before merging.

seabass-labrax:
Thank you! I'm looking forward to this change :)

PR #498, opened 2022-02-08
Make update-resources pull from latest stable release

### LINKED ISSUE #490: `make update-resources` should not pull from master
Issue raised in #489

mxmehl:
Fixes #490

I use a different Github API than @carmenbianca introduced in #489 that directly outputs the latest stable release (no draft or pre-release) and shows the connected tag.

~~Also bumps the SPDX license list data to 3.16, released today.~~

### DIFF
--- a/.github/workflows/license_list_up_to_date.py
+++ b/.github/workflows/license_list_up_to_date.py
@@ -8,30 +8,43 @@ up-to-date.
 For convenience, also overwrite the files.
 """
 
+import argparse
 import json
 import sys
 import urllib.request
 from pathlib import Path
 
-from packaging import version
-
-API_URL = "https://api.github.com/repos/spdx/license-list-data/tags"
+API_URL = "https://api.github.com/repos/spdx/license-list-data/releases/latest"
 URLS = {
+    # pylint: disable=line-too-long
     "exceptions.json": "https://raw.githubusercontent.com/spdx/license-list-data/{tag}/json/exceptions.json",
     "licenses.json": "https://raw.githubusercontent.com/spdx/license-list-data/{tag}/json/licenses.json",
 }
 
 
+# Fetch arguments
+parser = argparse.ArgumentParser(
+    description="Check and update included SPDX licenses and exceptions"
+)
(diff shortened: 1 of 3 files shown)


carmenbianca:
1. The use of the different API should be backported to the github workflow
2. Ideally, both the github workflow and the Makefile would simply run the same Python script.

carmenbianca:
> while the license_list_up_to_date.py script does additional things beyond just downloading the latest files

It really doesn't :) Most of the script is scaffolding, and the only additional functionality is checking whether the files are already up-to-date or not.

> My rationale was to make make update-resources independent from Python and its libraries to be as portable as possible. But since this process is supposed to be only run by project maintainers anyway, we can also higher the threshold.

A lot of stuff in the Makefile already has a lot of dev requirements (e.g., `make black`). The download script discussed here uses standard libraries only.

The reason I'm saying to get the workflow and the Makefile to use the same Python script is to be as sure as possible that there won't be any discrepancy between them.

linozen:
I rebased this onto master and made it so that the same script is used in the workflow and the Makefile. Also, I removed an unnecessary step in the workflow. AFAIC this is ready to be merged.

linozen (approved):

floriansnow:
@linozen Did something go wrong with the rebase or it Github unable to treat the rebase properly?  It currently shows a huge change that I know is not all part of this PR.  So a review is currently tricky.

linozen:
@floriansnow Sorry, my git-fu sucks sometimes. I rebased properly (had to force-push unfortunately), and properly merged 'master' into this PR. Now, it should be more straightforward to review.

mxmehl:
Wait, somehow my changed in the Makefile got overriden. Actually, I used a different API for fetching the latest stable release.

And yes, I see the point that this could be incorporated in `license_list_up_to_date.py` to ensure that the CI run and the Makefile step do the same things.

For reference, the latest stable release could best be determined via `curl https://api.github.com/repos/spdx/license-list-data/releases/latest | grep tag_name | cut -d'"' -f4`

PR #499, opened 2022-02-08
Separate directory for Dockerfiles

mxmehl:
It may happen that we will support more Dockerfile variations in the future. To no clutter the root directory, I moved them to a separate directory.

I am not entirely sure whether the process of tag releases will work this way, but we can find out after merge and the :dev tag.


linozen:
This change makes a lot of sense to me and can be merged AFAIC :+1:

PR #500, opened 2022-02-08
Add script in documentation for requested use-cases

nicorikken:
Add a section in the documentation with scripts to help with getting the information for the addheader command.

Provides some examples related to #297 and #349.

More work is necessary to suggest copyright on a per-file basis and to provide author details for reuse users wanting to add headers for their own contributions.

To be explicit about the dual-licensed copyright the SPDX snippet headers are added. These are not yet supported by SPDX but are intended for SPDX 3.0 https://github.com/spdx/spdx-spec/pull/464 If the SPDX specification turns out differently, these headers can be modified appropriately.

Signed-off-by: Nico Rikken <nico.rikken@fsfe.org>

### DIFF
--- a/docs/scripts.rst
+++ b/docs/scripts.rst
@@ -7,13 +7,13 @@
 Helper scripts
 ==============
 
-This section contains scripts and snippets to help with the usage of reuse.
-Reuse already has a lot of functionality builtin, but there are still cases where reuse could use support from some external scripts.
+This section contains scripts and snippets to help with the usage of REUSE.
+REUSE already has a lot of functionality builtin, but there are still cases where REUSE could use support from some external scripts.
 A typical example is adding SPDX headers based on the information in the version control system.
 This collection of scripts and snippets offers help for such situations.
 
 .. warning::
-  These scripts help you run reuse-tool against your codebase.
+  These scripts help you run REUSE against your codebase.
   Automatically extracted information might not resemble the truth.
   The correctness of these scripts is not guaranteed.
   Use with caution and at your own risk.
--- a/docs/scripts.rst
+++ b/docs/scripts.rst
@@ -149,7 +149,7 @@ A sign-off annotation in a commit also contains author details that can be as va
 
 .. code-block:: console
 
-  $ git log --all | grep 'Signed-off-by\|Co-authored-by' | sort | uniq
+  $ git log --all | grep -i 'Signed-off-by\|Co-authored-by' | sort | uniq
       Co-authored-by: Ethel Morgan <eth@ethulhu.co.uk>
       Co-authored-by: max.mehl <max.mehl@fsfe.org>
       Co-authored-by: Max Mehl <max.mehl@fsfe.org>
--- a/docs/scripts.rst
+++ b/docs/scripts.rst
@@ -1,0 +1,164 @@
+..
+  SPDX-FileCopyrightText: 2022 Nico Rikken <nico.rikken@fsfe.org>
+
+  SPDX-License-Identifier: CC-BY-SA-4.0
+
+==============
+Helper scripts
+==============
+
+This section contains scripts and snippets to help with the usage of reuse.
+Reuse already has a lot of functionality built-in, but there are still cases where reuse could use support from some external scripts.
+A typical example is adding SPDX headers based on the information in the version control system.
+This collection of scripts and snippets offers help for such situations.
+
+.. warning::
+  These scripts help you run reuse-tool against your codebase.
+  Automatically extracted information might not resemble the truth.
+  The correctness of these scripts is not guaranteed.
+  Use with caution and at your own risk.
+
+******************************
+Starting point of the codebase
+******************************
+
+The first code contribution can be a worthwile date to include in the copyright annotation.
+
+First commit
+============
+
+Git log can show summaries of all commits.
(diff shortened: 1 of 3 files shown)


nicorikken:
As I noticed some mentions of Mercurial for version control, I can see this page be extended with mercurial-based examples, which can be displayed side-by-side using an extension like [sphinx-tabs](https://sphinx-tabs.readthedocs.io/en/latest/#code-tabs).

docs/scripts.rst
```diff
@@ -0,0 +1,194 @@
...
+  SPDX-FileCopyrightText: 2022 Nico Rikken <nico.rikken@fsfe.org>
+
+  SPDX-License-Identifier: CC-BY-SA-4.0
+
+==============
+Helper scripts
+==============
+
+This section contains scripts and snippets to help with the usage of reuse.
+Reuse already has a lot of functionality built-in, but there are still cases where reuse could use support from some external scripts.
```
mxmehl, line 11: Can we settle on one way to write REUSE? I'd opt for REUSE generally, or "REUSE helper tool" when we speak specifically about the tool.

docs/scripts.rst
```diff
@@ -0,0 +1,194 @@
...
+A sign-off annotation in a commit also contains author details that can be as valuable.
+
+.. TODO: improve this oneliner based on built-in Git options as documented in https://stackoverflow.com/a/41361273/12013233
+
+.. SPDX-SnippetBegin
+.. SPDX-Snippet-License-Identifier: CC0-1.0
+
+.. code-block:: console
+
+  $ git log --all | grep 'Signed-off-by\|Co-authored-by' | sort | uniq
```
mxmehl, line 152: May it make sense to make a `grep -i`? I could imagine that there are variations.

docs/scripts.rst
```diff
@@ -0,0 +1,194 @@
...
+  These scripts help you run reuse-tool against your codebase.
+  Automatically extracted information might not resemble the truth.
+  The correctness of these scripts is not guaranteed.
+  Use with caution and at your own risk.
+
+******************************
+Starting point of the codebase
+******************************
+
+The first code contribution can be a worthwile date to include in the copyright annotation.
```
mxmehl, line 25: worthwhile

docs/scripts.rst
```diff
@@ -0,0 +1,194 @@
...
+.. SPDX-SnippetEnd
+
+Year of first commit
+====================
+
+With a custom format just the year of the first commit can be displayed.
+This output is convenient for use in larger scripts.
+
+.. SPDX-SnippetBegin
+.. SPDX-Snippet-License-Identifier: CC0-1.0
```
mxmehl, line 52: It's not defined, but wouldn't a REUSE compliant snippet also require a copyright line?

mxmehl (commented):
Smaller comments, otherwise LGTM!

docs/scripts.rst
nicorikken: Great, I was looking for guidance on it while writing. i changed it.

docs/scripts.rst
nicorikken: I see that use-case, even though it is typically handled by automation and therefore properly capitalized. It wouldn't hurt to ignore case here.

docs/scripts.rst
nicorikken: I was thinking about it while writing, as you explicitly stated the copyright in your example in https://github.com/spdx/spdx-spec/pull/464 Yet if we have the same author for the scripts as the rest of the text, I don't see the need to duplicate the declaration. But that discussion might be better served for the SPDX repository. Also, the original author seems less relevant for a CC0 license.

I'm happy to change it if you still think it add value.

mxmehl: True, I am also split. I'd say we can leave it for now, and see how the SPDX spec and our implementation evolves.

mxmehl (approved):

linozen (approved):

floriansnow:
@nicorikken Where were we in regards to the discussion about whether this should go in a separate repo to more strongly signal that these scripts are not officially supported and may be out of date/sync with the REUSE tool at times?

mxmehl:
> @nicorikken Where were we in regards to the discussion about whether this should go in a separate repo to more strongly signal that these scripts are not officially supported and may be out of date/sync with the REUSE tool at times?

The last comment on this discussion in our chat came from me:

It's a tricky one and I can live with both. Initially, I was in favour of the separate repo to be able to push away responsibility and separate things clearly. But I realised that integrating it in the tool's documentation has a few attractive advantages:

* It's easy to find. We can assume people who are interested in advanced functionality will crawl through the docu first before raising an issue. In a separate repo, even if we announce it, this might be less likely.
* It's easier to maintain. Once we add a new functionality that was covered by a custom pipe earlier, it's more likely we also update the corresponding part in the scripts section.
* Setup-wise it is low-effort: we can use the readthedocs integration as before and don't need to populate another instance for this. Sure, we could keep it Markdown-only, but I also see the point of more advanced formatting features that Nico already brought in.

With the disclaimer in place, I could live with the potential downsides, and think they are outweighed by the advantages.

linozen:
> With the disclaimer in place, I could live with the potential downsides, and think they are outweighed by the advantages.

I agree with your analysis and would like to merge this if there are no other issues raised until the weekend. These scripts are really helpful, thanks @nicorikken

PR #509, opened 2022-04-06
Switch to poetry

### LINKED ISSUE #289: Fully migrate to `poetry`
This should more permanently deal with problems such as #288.

Super low priority.

linozen:
This should close #289. More generally, this is a bit step towards better reproducibility of both the development environment and the resulting binaries. This is the case as poetry captures the entire dependency tree of `reuse` in [`poetry.lock](https://github.com/fsfe/reuse-tool/blob/switch-to-poetry/poetry.lock).

For more information about why reproducibility matters, see the [docs of reproducible-builds.org](https://reproducible-builds.org/docs/.)

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -22,3 +22,7 @@ License: GPL-3.0-or-later
 Files: src/reuse.egg-info/*
 Copyright: 2017 Free Software Foundation Europe e.V. <https://fsfe.org>
 License: GPL-3.0-or-later
+
+Files: poetry.lock
+Copyright: 2017 Free Software Foundation Europe e.V. <https://fsfe.org>
+License: CCO-1.0
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,1743 @@
+[[package]]
+name = "alabaster"
+version = "0.7.12"
+description = "A configurable sidebar-enabled Sphinx theme"
+category = "dev"
+optional = false
+python-versions = "*"
+
+[[package]]
+name = "astroid"
+version = "2.9.3"
+description = "An abstract syntax tree for Python with inference support."
+category = "dev"
+optional = false
+python-versions = ">=3.6.2"
+
+[package.dependencies]
+lazy-object-proxy = ">=1.4.0"
+typed-ast = {version = ">=1.4.0,<2.0", markers = "implementation_name == \"cpython\" and python_version < \"3.8\""}
+typing-extensions = {version = ">=3.10", markers = "python_version < \"3.10\""}
+wrapt = ">=1.11,<1.14"
+
+[[package]]
+name = "atomicwrites"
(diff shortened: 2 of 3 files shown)


mxmehl (approved):

carmenbianca:
This is awesome! I need to find some time to review this.

mxmehl:
@carmenbianca did you already find time to review this or is this good to go for you? It's a bit of a blocker for further steps :)

carmenbianca (approved):
Finally got the chance to review this. I can't speak to the Nix scrips—the syntax is completely foreign to me, and all my exposure to Nix is a single super interesting talk.

The poetry config makes perfect sense to me.

Some notes:

- Basic usage should probably be documented. `CONTRIBUTING.md` is probably a good destination for this. As a matter of fact, there's some stuff in the README that should probably be moved to that destination, but we can delay that.
- There is currently no feature parity between the poetry config and the setuptools config.
  - Translations aren't included in the built package.
  - The long description is not included in the built package.
  - The license files are not included in the built package.

However, none of these things are blockers if we want to maintain this as a parallel project until we deprecate the setuptools methods. If we _do_ deprecate the setuptools methods, I think it'd be nice to get rid of the custom Makefile as well. It's a bit clunky and awkward here and there.

LGTM, thanks a lot @linozen, and sorry for the long wait!


edit: When this is merged, it's a good idea to create an issue for feature paritry.

floriansnow:
@linozen LGTM! Is the plan to completely switch to this as a build system at some point or does it mean we have to maintain two separate setups forever?  Sorry if that is a silly question, but I'm not familiar enough with this new part yet.

floriansnow (approved):

linozen:
> @linozen LGTM! Is the plan to completely switch to this as a build system at some point or does it mean we have to maintain two separate setups forever? Sorry if that is a silly question, but I'm not familiar enough with this new part yet.

Thanks for approving. So, I think it would be beneficial to switch to `poetry` in the medium term. However, there is still some work left to do until we can abandon the old build/publish system based on `setup.py`. After this is merged, I'll create a new issue for this migration.

The Nix* stuff is mainly there for me so I can develop effectively. Furthermore, it also allows other Nix users to spin up a development environment in no time. More generally, IMHO efforts like Nix or Guix that focus on reproducibility,  declarativeness and reliability to be the future of packaging _any_ application. For now, I'd just be very happy if the two files ending in `.nix` could live on master so I don't have to worry about keeping them in sync across my development machines. And as I mentioned, they are not completely without use for a potentially relevant subset of outside contributors.

floriansnow:
> Furthermore, it also allows other Nix users to spin up a development environment in no time.
I find Nix, especially the language, still a bit scary, but I can see that I need to find the time to try it, even if not as my main system. :)

PR #513, opened 2022-05-09
Update SPDX licenses/exceptions lists to 3.17

mxmehl:
3.17 has been released today.

I used the new routine in the Makefile to update the resources to test it in practice.


mxmehl:
This should be a no-brainer. However, please do not merge before tomorrow 09:00 UTC as I would like to see whether the license checker workflow succeeds (or rather fails as expected).

PR #515, opened 2022-05-10
Added support for Vimscript files (`.vim`)

### LINKED ISSUE #502: Add vimscript comment syntax (" comment)
Comment syntax from vim docs:
```
*:quote* *:comment*
'"' at the start of a line causes the whole line to be ignored.  '"'
after a command causes the rest of the line to be ignored.  This can be used
to add comments.  Example:
    :set ai     "set 'autoindent' option
```

Possibly something like this (in https://github.com/fsfe/reuse-tool/blob/master/src/reuse/_comment.py)
```
class VimCommentStyle(CommentStyle):
    """Vim comment style."""

    _shorthand = "vim"

    SINGLE_LINE = "\""
    INDENT_AFTER_SINGLE = " "
```

and mapping the extension like this:
```
FILENAME_COMMENT_STYLE_MAP = {
     [...]
    ".vim": VimCommentStyle,
}
```

For extensionless files, I suppose the `--style` option can be used :-).

linozen:
This should close #502.

### DIFF
--- a/src/reuse/_comment.py
+++ b/src/reuse/_comment.py
@@ -445,6 +445,15 @@ class UncommentableCommentStyle(EmptyCommentStyle):
     """
 
 
+class VimCommentStyle(CommentStyle):
+    """Vim(Script|Config) style."""
+
+    _shorthand = "vim"
+
+    SINGLE_LINE = '"'
+    INDENT_AFTER_SINGLE = " "
+
+
 #: A map of (common) file extensions against comment types.
 EXTENSION_COMMENT_STYLE_MAP = {
     ".adb": HaskellCommentStyle,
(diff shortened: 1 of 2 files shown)


src/reuse/_comment.py
```diff
@@ -678,6 +688,7 @@ class UncommentableCommentStyle(EmptyCommentStyle):
     "ROOT": MlCommentStyle,
     "setup.cfg": PythonCommentStyle,
     "sonar-project.properties": PythonCommentStyle,
+    ".vimrc": VimCommentStyle,
```
mxmehl, line 691: Nitpick: the sorting doesn't look right. The dotfiles are at the beginning of the list.

mxmehl (approved):
Please fix the sorting, otherwise LGTM and tested.

src/reuse/_comment.py
linozen: Sure thing :+1:

PR #517, opened 2022-05-13
Fix copyright detection

### LINKED ISSUE #440: Cannot detect copyright
I tested lint mode with the below file. 
And it cannot detect the copyright even though the header of the file includes the copyright text.
If 'copyright' and '(c)' are written no leaving space, it is not recognized as copyright.
Please fix it.
```
/*
 * ============================================================================
 * @Project       : test project
 * @Version       : v1.0.0
 *
 * Copyright(c) 2012 TEST All rights reserved
 */

```

linozen:
This should fix #440 by simply expanding the copyright detection mechanism.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,9 @@ The versions follow [semantic versioning](https://semver.org).
 
 - Updated SPDX license list to 3.17. (#513)
 
+- The copyright detection mechanism now silently accepts the following strings:
+  `Copyright(c)` and `Copyright(C)`. (#440)
+
 ### Deprecated
 
 - Deprecated `--explicit-license` in favour of `--force-dot-license`.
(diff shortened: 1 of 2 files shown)


mxmehl:
It works, but I am not so much of a fan of this solutions as it makes the no-space styles selectable:
```
$ reuse addheader -h
  --copyright-style {spdx,spdx-symbol,string,string-c,string-c-lower,string-c-no-space,string-c-lower-no-space,string-symbol,symbol}
```

IMHO it would make more sense to silently understand these styles but not support adding them like this (and thereby also confuse users). Only the ones we also list [here](https://reuse.readthedocs.io/en/stable/usage.html#addheader) should be offered.

linozen:
> It works, but I am not so much of a fan of this solutions as it makes the no-space styles selectable:

I was not aware of this. Now, I am :tada:. In this case, simply matchting the whitespace before `(c)` or `(C)` should do it then, no?

mxmehl:
Yes, that should do it, thank you! Would you please add a changelog?

mxmehl (approved):

PR #525, opened 2022-05-24
Sanitize xargs input in scripts documentation

nicorikken:
The xargs command listed in the scripts section of the documentation
uses the unsanitized input from the `git diff` command. This breaks when
files have spaces in their filename. By using the `-I` flag the input
can be quoted.


carmenbianca (approved):
LGTM

PR #529, opened 2022-05-31
Add AIDL and Proguard comment support

FestplattenSchnitzel:


FestplattenSchnitzel:
Thank you for the review!

I removed the `.pro` line and added a changelog entry.

mxmehl (approved):
Thanks for fixing and the changelog!

PR #532, opened 2022-06-01
Remove latesttag workflow

mxmehl:
The workflow tries to fetch the `latest` tag which does not exist any more.

The script also works with just fetching the main branch (currently `master`).

On the other hand we could also consider removing this workflow as we don't need to manually confirm this moving tag. It would just warn us if we somehow forgot to release a new version on PyPI when we released one in Git.


carmenbianca:
This can safely be removed I feel.

mxmehl:
I have the same feeling. Updated this PR to remove it. Please check and review :)

carmenbianca (approved):

Issue #524, opened 2022-05-24
Support .pem files


nicorikken:
I encountered a case of a `.pem` certificate file in a repository. Typically `.pem` files are text based and contain a key or certificate. As the relevant content is to be defined inbetween tags like `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` no real comment style is needed, so long as the text is outside of these markers.

Based a discussion in another project https://github.com/aws/s2n-tls/issues/1136 I found an example where `#` comment prefixes are used in the `.pem` file. If we don't want any prefix, we could even use the base class CommentStyle, but perhaps it would be better to specify an explicit comment style like AllCommentStyle as EmptyCommentStyle is already taken for a different purpose.

Looking through GitHub I couldn't find any .pem file that included a SPDX header: https://github.com/search?p=1&q=extension%3A.pem+spdx&type=Code Which does make sense in a way, because it is configuration and mostly generated, so why bother. But REUSE forces us to be complete, so I think it is still relevant to support SPDX headers.

What is preferred for adding SPDX headers to a `.pem` file?

mxmehl:
As certificates are quite sensible and no standard seems to exists, I'd tend to marking them as uncommentable by default, which means that by default a .license file is created.

If someone feels brave and wants to add a comment, the tool would still detect it. It's quite similar to the SVG discussion.

floriansnow:
So I guess we can close this @nicorikken ?

mxmehl:
> So I guess we can close this @nicorikken ?

If we agree that .pem shall be uncommentable, we should add this extension to the Uncommentable list.

PR #539, opened 2022-06-09
Make .pem files uncommentable

linozen:
Is that all that's needed for closing #524?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -52,6 +52,8 @@ Contributors
 
 - Florian Snow <florian@familysnow.net>
 
+- Linus Sehn <linus@fsfe.org>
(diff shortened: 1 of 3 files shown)


nicorikken:
Technically yes, so thanks. But please add the administration to add the new syntax in the https://github.com/fsfe/reuse-tool/blob/master/CHANGELOG.md and please check your attribution is in place in https://github.com/fsfe/reuse-tool/blob/master/AUTHORS.rst and the SPDX headers of the file. Credit where credit is due! (and to be explicit about copyright, that as well)

nicorikken (approved):
Great!

carmenbianca (approved):

PR #548, opened 2022-06-25
Actually exit with returned status when invoking via `__main__.py`

### LINKED ISSUE #547: Exit status 0 when linting fails if run using `python -m reuse lint`
I've installed `reuse` package using `pip` and it works for me when running it using the installed entrypoint (`reuse` on PATH).

However, when using the `python -m` syntax to start an entrypoint it seems to swallow the errors and exit status 0 (OK) on failure.

```
$ python -m reuse lint
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
[...]

# SUMMARY
[...]

Unfortunately, your project is not compliant with version 3.0 of the REUSE Specification :-(
$ echo $?
0
```

```
$ reuse lint
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
[...]

# SUMMARY
[...]

Unfortunately, your project is not compliant with version 3.0 of the REUSE Specification :-(
$ echo $?
1
```

I need to use the `python -m [module]` syntax for situation where the entrypoints are not on PATH, e.g. as Task in VS Code; you get the path to the Python interpreter, but not an appropriate PATH. All other linters I use have do not suffer from this issue.

gertvdijk:
Fixes #547

### DIFF
--- a/src/reuse/__main__.py
+++ b/src/reuse/__main__.py
@@ -4,7 +4,10 @@
 
 """Entry module for reuse."""
 
+import sys
+
+
 if __name__ == "__main__":
     from ._main import main
 
-    main()
+    sys.exit(main())


wbolster (approved):

linozen (approved):
Thanks @gertvdijk for contributing to REUSE! It is highly appreciated.

For the other reviewers: [Here](https://docs.python.org/3/library/__main__.html#idiomatic-usage) is the reference for the idiomatic usage of `sys.exit(main())`

Issue #553, opened 2022-07-04
Support Apache Velocity template files


nicorikken:
I encountered `.vtl` and `.vm` files in the wild, which are template files for [Apache Velocity](https://velocity.apache.org/). More about the comment syntax in the [VTL Reference](https://velocity.apache.org/engine/devel/vtl-reference.html#comments).

Especially the single line comment is a difficult one, requiring some refactoring for support for the trailing `**`:
```
## This is a comment. **
```
The multi-line is easy to implement:
```
#*
  This is a multiline comment.
  This is the second line.
*#
```

I'll make a simple PR for the multi-line comment, leaving this issue as a reminder to work on a change for the comment handling to support trailing markers in single-line comments.

mxmehl:
Thanks, LGTM so far.

Regarding the trailing whitespace, two comments:
1. Does the syntax really require a whitespace?
2. The tool could learn to write an empty line when the line will just contain the whitespace.

nicorikken:
I did some more searching and I found that whitespace is not required, going by https://velocity.apache.org/engine/devel/user-guide.html#comments Let me copy the examples here:
```
## This is a single line comment.
```

```
This is text that is outside the multi-line comment.
Online visitors can see it.

#*
  Thus begins a multi-line comment. Online visitors won't
  see this text because the Velocity Templating Engine will
  ignore it.
*#

Here is text outside the multi-line comment; it is visible.
```
```
#**
  This is a VTL comment block and
  may be used to store such information
  as the document author and versioning
  information:
    @author John Doe
    @version 5
*#
```
I didn't know about the latter, but that might be most relevant as it is intended to document information about the document, just as SPDX headers do.

**New proposal**
So on second thought I think it makes sense to change it so we end up with:
```
#**
SPDX-FileCopyrightText: MyOrg

SPDX-License-Identifier: AGPL-3.0-or-later
*#
```
I can modify the pull request accordingly.

PR #554, opened 2022-07-04
Apache Velocity template language comment style

nicorikken:
Adds comment style VelocityCommentStyle for the template files of the
Apache Velocity project https://velocity.apache.org/ Most typical file
formats `.vst` and `.vm` for these files are assigned for this comment
style.

This PR relates to issue #553

The single line comments requires a more elaborate code change to
support trailing ` **`, so this change only contains the multi-line
format.


carmenbianca:
This will break after #593 is merged.

linozen:
After #593 is merged, we'd need to use the `MultiLineSegments` class to fix this issue.

PR #559, opened 2022-07-17
Remove unneeded exclude config for black

carmenbianca:
For some reason this wasn't working for me, and black started formatting everything in `.tox`. Removing this config fixed it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,19 +5,6 @@
 
 [tool.black]
 line-length = 80
-exclude = '''
-/(
-    \.git
-  | \.hg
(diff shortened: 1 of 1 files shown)


nicorikken:
This makes sense to me. However, running `poetry run black -v .` I see it also modifies some dubious directories on my system:

- `.eggs/setuptools_scm-7.0.5-py3.8.egg/setuptools_scm/
- `build/lib/reuse/`

However, I think both wouldn't hurt, as they represent the source code that is already properly formatted.

nicorikken (approved):

PR #560, opened 2022-07-18
Find license identifiers in comments with ASCII art frames

### LINKED ISSUE #343: Multiline block comments with extraneous characters after the MULTI_LINE[1] indicator throw out the C-style parser
LLVM has these beautiful box-comments in which the SPDX licensing info gets inserted, leading to this error:

`Could not parse 'Apache-2.0 WITH LLVM-exception                    *|'`

Note: the above has multiple spaces prior to `*|`, but Github ate them.

Example: 
```
/*==-- clang-c/BuildSystem.h - Utilities for use by build systems -*- C -*-===*\
|*                                                                            *|
|* Part of the LLVM Project, under the Apache License v2.0 with LLVM          *|
|* Exceptions.                                                                *|
|* See https://llvm.org/LICENSE.txt for license information.                  *|
|* SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception                    *|
|*                                                                            *|
|*===----------------------------------------------------------------------===*|
|*                                                                            *|
|* This header provides various utilities for use by build systems.           *|
|*                                                                            *|
\*===----------------------------------------------------------------------===*/
```

I'd exclude the regex ` *\*[|/]$`, but that's not how the code works in:
https://github.com/fsfe/reuse-tool/blob/e7040863a265db9b2314a2f29b3fcfd59f5cae51/src/reuse/_comment.py#L172-L182

I'd add a block that, when `cls.MULTI_LINE[1]` is non-empty, runs a `possible_line.rsplit(cls.MULTI_LINE[1], maxsplit=1)[0]` and instead try to parse that. I think this would >not< require changes later lines 191 and 196, but I think the coupling between these blocks is a bit fragile:
https://github.com/fsfe/reuse-tool/blob/e7040863a265db9b2314a2f29b3fcfd59f5cae51/src/reuse/_comment.py#L191-L196

To address this use case of extraneous characters, I think the best approach is to drop off anything after the expected `MULTI_LINE[1]`. If however the character isn't the expected one (say, the `*` is missing from this example), the comment parser breaks in the same way as today. I think that's no regression but that it's just a fragile parser; thoughts on that are welcome, but I'd say I'm proposing a strict improvement, merely incomplete.

Thanks in advance!

pietroalbini:
The Rust project is working to [adopt REUSE to annotate licenses](https://github.com/rust-lang/compiler-team/issues/519), but when working on the initial implementation I stumbled on #343. The Rust repository has LLVM as one of its submodules, and REUSE currently errors out on some of the LLVM comments:

```
reuse._util - ERROR - Could not parse 'Apache-2.0 WITH LLVM-exception                    *|'
reuse.project - ERROR - 'src/llvm-project/lldb/source/Plugins/Plugins.def.in' holds an SPDX expression that cannot be parsed, skipping the file                                                                      
```

As #343 correctly pointed out, the problem is that LLVM uses ASCII art "frames" for those code comments, like:

```c
/***********************************************************\
|* SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception *|
\***********************************************************/
```

The solution I came up with is to implement generic handling for these kinds of ASCII art, even in cases where the multiline delimiter (in this case `*`) is not present. When there are some non-whitespace chars before `SPDX-License-Identifier`, the new code tries to strip the *reverse* of them from the end of the line. That correctly handles LLVM comments, but also any other ASCII art frame that's symmetric.

Fixes #343

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,6 +56,8 @@ Contributors
 
 - Yaman Qalieh
 
+- Pietro Albini <pietro.albini@ferrous-systems.com>
+
 Translators
 -----------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@ The versions follow [semantic versioning](https://semver.org).
 ### Fixed
 
 - Sanitize xargs input in scripts documentation
+- License identifiers in comments with symmetrical ASCII art frames are now
+  properly detected (#560)
 
 ### Security
 
--- a/src/reuse/_util.py
+++ b/src/reuse/_util.py
@@ -4,6 +4,7 @@
 # SPDX-FileCopyrightText: 2022 Nico Rikken <nico.rikken@fsfe.org>
 # SPDX-FileCopyrightText: 2022 Florian Snow <florian@familysnow.net>
 # SPDX-FileCopyrightText: 2022 Carmen Bianca Bakker <carmenbianca@fsfe.org>
+# SPDX-FileCopyrightText: 2022 Pietro Albini <pietro.albini@ferrous-systems.com>
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -22,7 +23,7 @@ from gettext import gettext as _
 from hashlib import sha1
 from os import PathLike
 from pathlib import Path
-from typing import BinaryIO, List, Optional, Set
+from typing import BinaryIO, Iterator, List, Optional, Set
 
 from boolean.boolean import Expression, ParseError
 from debian.copyright import Copyright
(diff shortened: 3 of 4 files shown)


carmenbianca (approved):
Looks good to me. I'm not sure if the case this covers is a good _general case_, but we can cross that bridge when someone else's ASCII art breaks.

Thanks for the PR! Please make sure to add a change log entry, and add yourself to AUTHORS.rst if you want.

pietroalbini:
Thanks for the quick review! I should've addressed everything.

mxmehl (approved):
Thanks, that's a clever approach a nice first step!

pietroalbini:
Hey all, thanks for the reviews! What are the next steps to get this merged and released?

PR #563, opened 2022-07-22
add c and c-lower copyright-style

vladh:
"Copyright" is English-specific, while the copyright symbol can be difficult to type. These are good reasons to use `(C)` or `(c)`, and I've seen many projects use these. This pull request adds support for `(C)` as style `c` and `(c)` as style `c-lower`.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -74,6 +74,8 @@ With the argument ``--copyright-style`` it is possible to change the default
 
 .. code-block::
 
+  c:              (C) <year> <statement>
+  c-lower:        (c) <year> <statement>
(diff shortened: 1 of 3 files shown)


carmenbianca:
I don't think that this PR can be merged until the specification is updated to reflect that `(C)` is a valid way to start a copyright statement.

I don't really oppose this proposal.

vladh:
Got it! I sent a spec PR: https://github.com/fsfe/reuse-docs/pull/108

linozen:
As discussed in our coordination meeting this would probably lead to too many false positives.

PR #565, opened 2022-07-26
feat: ansible.cfg comment style support

nicorikken:
Support the INI format syntax for `ansible.cfg` files. This file is used
to configure the Ansible software and can reside in multiple locations.
The INI format uses the hash sign (#) for comments that occupy an entire
line. The PythonCommentStyle is chosen to match this comment style.

More documentation about the Ansible Configuration Settings file is
available in the official documentation:
https://docs.ansible.com/ansible/latest/reference_appendices/config.html

### DIFF
--- a/src/reuse/_comment.py
+++ b/src/reuse/_comment.py
@@ -661,6 +661,7 @@ FILENAME_COMMENT_STYLE_MAP = {
     ".Rprofile": PythonCommentStyle,
     ".vimrc": VimCommentStyle,
     ".yarnrc": PythonCommentStyle,
+    "ansible.cfg": PythonCommentStyle,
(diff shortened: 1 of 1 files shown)


mxmehl (approved):
Oh god, I stumbled to often over this. Thanks :)

mxmehl:
May I ask you to add the changelog?

nicorikken:
@mxmehl via chat we concluded that it would be easier to create the changelog as part of the release process, to avoid merge conflicts due to changelog entries. Happy to do it, but as already some changes are missing from the changelog and we discussed it, I think we should do it later when a new release is drafted.

nicorikken:
> Oh god, I stumbled to often over this. Thanks :)

I try to add a comment style for every file or filetype I encounter to avoid having to think about it again.

PR #573, opened 2022-08-20
Add style for php 'blade' templates

raboof:


nicorikken:
Thanks @raboof! I'm afraid we will encounter more double extensions in the future and then this might be helpful. Do you have some examples for that comment style, ideally where a license header is used?

I think the plural array name is confusing. How about renaming `EXTENSIONS_COMMENT_STYLE_MAP_LOWERCASE` to `MULTIPLE_EXTENSION_COMMENT_STYLE_MAP_LOWERCASE` It is longer of course.

Also I think this needs a test to check precedence to avoid breaking things when refactoring in the future.

raboof:
> Do you have some examples for that comment style, ideally where a license header is used?

Yes, at https://codeberg.org/eicas/website/src/branch/add-mit-license/resources/views/home.blade.php

> I think the plural array name is confusing. How about renaming `EXTENSIONS_COMMENT_STYLE_MAP_LOWERCASE` to `MULTIPLE_EXTENSION_COMMENT_STYLE_MAP_LOWERCASE` It is longer of course.

Sure

> Also I think this needs a test to check precedence to avoid breaking things when refactoring in the future.

Makes sense. Fair warning: I might not get to this soon, though.

raboof:
Sorry to let this linger, I've rebased, fixed conflicts and added a test.

carmenbianca:
It took a while, but I've slightly refactored this, and it's ready to merge :) Thank you @raboof

raboof:
Thanks!

PR #574, opened 2022-08-21
Allow overriding options for pre-commit hook

marktsuchida:
Meson subprojects are (surprisingly, at least to me) ignored by default even if not a git submodule, so being able to pass `--include-meson-subprojects` is essential for being able to use `reuse`, including as a pre-commit hook. This is not currently possible on the user's end.

By moving `lint` out of `entry` and into `args`, users of the hook can override the args in order to pass options like `--include-submodules` or `--include-meson-subprojects` (these options need to be given before the subcommand `lint`).

Any users that currently override args will need to add `lint` to their `args` when they update the hook version. Such users are likely rare/nonexistent because `reuse lint` currently has no options that are useful in a pre-commit hook context.

The default behavior (with no `args` in users' `.pre-commit-config.yaml`) does not change: it just runs `reuse lint`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,7 +4,8 @@
 
 - id: reuse
   name: reuse
-  entry: reuse lint
+  entry: reuse
+  args: ["lint"]
   language: python
   pass_filenames: false
   description:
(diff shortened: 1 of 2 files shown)


mxmehl:
Thanks, and LGTM. However, I am a rare user of the .pre-commit file. Maybe @carmenbianca wants to review?

carmenbianca (approved):
LGTM. We reviewed this during the monthly call, and this makes perfect sense.

PR #584, opened 2022-09-01
Add support for XQuery file comments and extensions

### LINKED ISSUE #583: support xquery comment syntax
While #177 suggests plugin support for obscure comment styles, the issue seems to have induced no attention and was not yet assigned a milestone. Thus, I suggest to implement the missing syntax the old fashioned way. A pull request with the suggested (minor) code changes will follow and I hope the feature can then soon be introduced.

stexandev:
XQuery only supports multiline comments with a syntax currently
unsupported by the reuse-tool, see
<https://www.w3.org/TR/xquery-31/#comments>.

Close #583

### DIFF
--- a/src/reuse/_comment.py
+++ b/src/reuse/_comment.py
@@ -7,6 +7,7 @@
 # SPDX-FileCopyrightText: 2021 Matija Šuklje <matija@suklje.name>
 # SPDX-FileCopyrightText: 2022 Florian Snow <florian@familysnow.net>
 # SPDX-FileCopyrightText: 2022 Nico Rikken <nico.rikken@fsfe.org>
+# SPDX-FileCopyrightText: 2022 Stefan Hynek <stefan.hynek@uni-goettingen.de>
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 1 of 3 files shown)


src/reuse/_comment.py
```diff
@@ -447,6 +448,17 @@ class VimCommentStyle(CommentStyle):
     INDENT_AFTER_SINGLE = " "
 
 
+class XQueryCommentStyle(CommentStyle):
+    """XQuery comment style."""
+
+    _shorthand = "xq"
```
carmenbianca, line 454: I would call this `xquery`

carmenbianca (changes requested):
Thanks for the contribution @stexandev !

A small nitpick.

Also, before this is merged, there needs to be a change log entry, and I believe there's somewhere in the documentation that needs to be updated as well.

src/reuse/_comment.py
stexandev: thank you for your feedback, @carmenbianca; additional commit(s) incoming!

stexandev:
> and I believe there's somewhere in the documentation that needs to be updated as well.

I could not find any place in the documentation where an update would have been necessary.

mxmehl (approved):
That should be alright, thanks! The first line in the Changelog wouldn't have been necessary though, but we tidy up the changelog before the next release anyway.

carmenbianca:
Merged in #610 

Thank you @stexandev !

Issue #516, opened 2022-05-13
`reuse addheader` cannot deal with a switch from single-line to multi-line comments


linozen:
Let's say you have a file `test.html` that already contains licensing or copyright info:

```html
<!-- SPDX-FileCopyrightText: 2017 Jane Doe -->
<!-- SPDX-License-Identifier: GPL-3.0-or-later -->

<p>Hello, world!</p>
```

if you then run reuse `addheader --copyright "Another Great Person" test.html`, you encounter an error.

The command works without error if the file looks like this:
```html
<!--
SPDX-FileCopyrightText: 2017 Jane Doe
SPDX-FileCopyrightText: 2022 Free Software Foundation Europe e.V. <https://fsfe.org>
-->

<p>Hello, world!</p>
```

carmenbianca:
This issue is a little broader. In fact, it's broken for a case that used to work, but apparently we never tested for?

It breaks on this file:

```
<!-- SPDX-FileCopyrightText: 2017 Jane Doe -->

<p>Hello, world!</p>
```

With a little tweak to the code, this is the traceback:

```
$ reuse addheader --license MIT test2.html
  warnings.warn("urllib3 ({}) or chardet ({})/charset_normalizer ({}) doesn't match a supported "
reuse.header - WARNING - 'reuse addheader' has been deprecated in favour of 'reuse annotate'
Traceback (most recent call last):
  File "/var/home/coopiteasy/.virtualenvs/reuse/bin/reuse", line 33, in <module>
    sys.exit(load_entry_point('reuse', 'console_scripts', 'reuse')())
  File "/var/home/coopiteasy/Projets/reuse-tool/src/reuse/_main.py", line 283, in main
    return parsed_args.func(parsed_args, project, out)
  File "/var/home/coopiteasy/Projets/reuse-tool/src/reuse/header.py", line 789, in run
    result += _add_header_to_file(
  File "/var/home/coopiteasy/Projets/reuse-tool/src/reuse/header.py", line 527, in _add_header_to_file
    raise e
  File "/var/home/coopiteasy/Projets/reuse-tool/src/reuse/header.py", line 507, in _add_header_to_file
    output = find_and_replace_header(
  File "/var/home/coopiteasy/Projets/reuse-tool/src/reuse/header.py", line 306, in find_and_replace_header
    header = create_header(
  File "/var/home/coopiteasy/Projets/reuse-tool/src/reuse/header.py", line 181, in create_header
[...]

PR #593, opened 2022-09-22
Fix sameline multiline

carmenbianca:
Related to #516, but doesn't really address it.

Fixes error when parsing

```
<!-- SPDX-FileCopyrightText: 2017 Jane Doe -->

<p>Hello, world!</p>
```


carmenbianca:
Merging because it's a blocker for #602

PR #594, opened 2022-09-30
Add MacPorts installation instructions

ryandesign:
You can [install reuse with MacPorts](https://ports.macports.org/port/reuse/)! This PR adds that information to the installation section of the readme.


carmenbianca:
That's awesome!

carmenbianca (approved):

PR #595, opened 2022-10-02
add Zsh completion

tastytea:
This adds completion for Zsh. It should be installed into `/usr/share/zsh/site-functions` but i can't figure out how to do that with this build system. 😅

<details> <summary>screenshots</summary>

reuse_zsh-completion
reuse_zsh-completion_licenses

</details>

### DIFF
--- a/completion/_reuse
+++ b/completion/_reuse
@@ -1,0 +1,148 @@
+#compdef reuse
+
+# SPDX-FileCopyrightText: 2022 tastytea <tastytea@tastytea.de>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+local curcontext=${curcontext} state state_descr line ret=1
+
+function _reuse_licenses() {
+    local -a licenses
+    OLDIFS=${IFS}
+    IFS=$'\n'
+    for line in $(reuse supported-licenses 2>&-); do
+        # remove everything before and including the first whitespace and
+        # everything after and including the first http
+        desc=${${line#*[[:blank:]]}%[[:blank:]]http*}
+        # trim whitespace
+        [[ ${desc} =~ '^\s*(\S.*\S)\s*$' ]] && desc=$match[1] || desc=''
+        licenses+=(${line%% *}:${desc})
+    done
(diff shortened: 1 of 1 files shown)


tastytea:
There is [pycomplete](https://github.com/frostming/pycomplete) (`pycomplete 'reuse._main:parser()' zsh`), but the result needs fixing and doesn't have license selection. It adds the long `description` to the short help text and doesn't add a help text to some subcommands for some reason.

<details> <summary>screenshot</summary>

grafik

</details>

There is [pyzshcomplete](https://github.com/dan1994/pyzshcomplete) which works by integrating it into the source code. I could not get it to work, probably because i don't know python and was just guessing where to put the code.

nicorikken:
The pyzshcomplete readme describes how the Zsh completion is generated on the fly. We could already do it at 'buildtime' perhaps in Python using the argument structure? It is a nice feature for improved usability.

linozen:
I added a meta issue to track completions for popular shells: #629

sgaist:
Hi,

One thing I have seen with several tools is that they have a "completion" subcommand that will output the completion file content for the selected shell on stdout. That way you do not require installation in a system specific (and possibly out of reach for the user) folder.

I think that would be a better alternative as it will allow users to make use of it with for example [oh my zsh ](https://ohmyz.sh/).

DaveMcEwan:
Having a completion sub-command also makes it easier to maintain support for other shells. In a Rust project [I work on](https://github.com/dalance/svlint), there's an argument `--shell-completion` which takes the name of any shell supported by [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/shells/index.html), i.e. Bash, Elvish, Fish, PowerShell, and Zsh. I think the installation process is too user-specific so just giving them a simple way to generate a completion script is a pragmatic compromise.
Mind you, the CLI of reuse-tool is so small and neat (thank you!) that this feature wouldn't really improve anybody's workflows in a significant way :p

floriansnow:
Closing this because #1076 is already merged.

PR #603, opened 2022-10-12
Enable tests using Python 3.11

carmenbianca:
Signed-off-by: Carmen Bianca Bakker <carmenbianca@fsfe.org>


carmenbianca:
It works! Let's merge this.

PR #606, opened 2022-10-12
Support deprecated licenses for download

### LINKED ISSUE #186: `reuse download $DEPRECATED` gives unexpected result
```bash
$ reuse download GPL-3.0                                                                                           
Error: Failed to download license. Is your internet connection working?
```

Oops.

Same goes for init.py.

carmenbianca:
Fixes #186


nicorikken (approved):
Great!

mxmehl (approved):

carmenbianca:
Can't be merged; needs to use urllib instead of requests.

floriansnow:
> Can't be merged; needs to use urllib instead of requests.

Would you please elaborate a bit about that problem? The code looks good to me.

Also, in order to fix #186 completely, should we perhaps only issue the message about a broken connection in case we can't download any licenses? We could even try to access a known-good license as a check.

carmenbianca:
> Would you please elaborate a bit about that problem? The code looks good to me.

After this PR was created, we merged a PR that removed the (direct) dependency on `requests`. The test code currently mocks/monkeypatches the requests library. We'll have to rewrite that line of code to use urllib instead.

> Also, in order to fix https://github.com/fsfe/reuse-tool/issues/186 completely, should we perhaps only issue the message about a broken connection in case we can't download any licenses? We could even try to access a known-good license as a check.

I don't know. It would be nice to improve the error handling and/or error message here, but I'm more inclined to do that as a separate issue/PR.

floriansnow:
Is there still interest in these changes?

PR #611, opened 2022-10-12
Move nix files out of project root

carmenbianca:
Ping @linozen


carmenbianca:
Nix files have to be in the project root. Let's close this. Maybe we can remove/alter the Nix files later.

PR #623, opened 2022-11-10
Populate the `LicenseConcluded` field in the SPDX report

### LINKED ISSUE #586: reuse spdx: also output the full `SPDX-License-Identifier`
The output of `reuse spdx` has the `LicenseInfoInFile `field.
However, with this field it is not always possible to determine the original value of `SPDX-License-Identifier`

For example if the input file has 
```
// Copyright (C) 2016 Paul Lemire <paul.lemire350@gmail.com>
// SPDX-License-Identifier: LicenseRef-Qt-Commercial OR GPL-3.0-only WITH Qt-GPL-exception-1.0
```
The output will be
```
FileName: ./qt3d/tests/auto/input/utils/tst_utils.cpp
SPDXID: SPDXRef-ec9701a4a6e83e5c346e9f4377005153
FileChecksum: SHA1: 78c032c19e50ba890da9609caca80ca3edea1166
LicenseConcluded: NOASSERTION
LicenseInfoInFile: GPL-3.0-only
LicenseInfoInFile: LicenseRef-Qt-Commercial
LicenseInfoInFile: Qt-GPL-exception-1.0
FileCopyrightText: <text>Copyright (C) 2016 Paul Lemire <paul.lemire350@gmail.com></text>
```

Which is not sufficient to determine the `SPDX-License-Identifier`

This information would be useful when using the `reuse spdx` output as input of another script (for example to build a DEP5 file)

pietroalbini:
This PR fixes #586 by adding the `--add-license-concluded` flag to the `reuse spdx` command.

When that flag is provided, REUSE will now calculate the `LicenseConcluded` field in the SPDX report rather than setting it to `NOASSERT`, by `AND`ing all the expressions related to the file. This will allow tools consuming the SPDX report to know the actual licensing of each file, rather than having to rely on the lossy/incorrect `LicenseInfoInFile` fields.

Passing the flag will result in a slightly noticeable slowdown in large repositories: on `rust-lang/rust`, without the flag `reuse spdx` takes 4.13s, while with the flag it takes 4.72s.

To ensure compliance with the SPDX spec, this also adds the `--creator-person` and `--creator-organization` flags to `reuse spdx`. Optional by default, at least one of the flags will be required when passing `--add-license-concluded`, as [requested in the issue](https://github.com/fsfe/reuse-tool/issues/586#issuecomment-1310425706).

~~Since there is no consensus yet in #586 whether to make this the default behavior or not, populating the field and gating that behind a CLI flag are in two different commits, so if you all decide to populate `LicenseConcluded` by default we can simply drop the last commit. This PR does not change `LicenseInfoInFile` in any way.~~

Also, the first commit makes running tests bearable on my machine, since I configured git to require signing all commits with my gpg key, which requires a physical touch on my YubiKey every time a commit is created :sweat_smile:


tenzap:
Will it work also when there are exceptions like in

`SPDX-License-Identifier: LicenseRef-Qt-Commercial OR GPL-3.0-only WITH Qt-GPL-exception-1.0`

pietroalbini:
Yes, this also works when there are exceptions! To be extra sure, I added an exception to the new test I added :slightly_smiling_face:

pietroalbini:
Added `reuse spdx --creator-person=NAME --creator-organization=NAME` as requested in https://github.com/fsfe/reuse-tool/issues/586#issuecomment-1310425706. The command will continue working without the flags, but at least one of them will be required when passing `--add-license-concluded`.

alpianon:
hi @pietroalbini we at [Oniro Project](https://docs.oniroproject.org/en/latest/) just realized that we would really need this functionality, it's great that you are already developing it, thanks! :)

FYI I'm proposing to add another spdx-related feature (support for comments, conversion of comments from dep5 to spdx, and possibly other comment-related features), see #625 . If you have any idea or feedback on that, feel free to comment there

mxmehl (approved):

mxmehl:
Great work, thanks for implementing all of @silverhook's suggestions!

src/reuse/report.py
```diff
@@ -119,8 +129,10 @@ def bill_of_materials(self) -> str:
 
         # Author
         # TODO: Fix Person and Organization
```
nicorikken, line 131: We can lose the todo now as well?

```suggestion
```

src/reuse/spdx.py
```diff
@@ -22,10 +23,47 @@ def add_arguments(parser) -> None:
     parser.add_argument(
         "--output", "-o", dest="file", action="store", type=PathType("w")
     )
+    parser.add_argument(
+        "--add-license-concluded",
+        action="store_true",
+        help=_(
+            "Populate the LicenseConcluded field. Note that REUSE cannot "
```
nicorikken, line 30: Reuse should be lowercase instead of capitalized I guess as we're talking about the tool, not the spec?
```suggestion
            "Populate the LicenseConcluded field. Note that reuse cannot "
```

nicorikken (approved):
2 small suggestions but overall great work! I can see this coming in handy in many cases, thanks for contributing it.

nicorikken:
:warning: Please don't merge this PR just yet, @carmenbianca has added some improvements on top if this work at https://github.com/fsfe/reuse-tool/pull/690

pietroalbini:
@nicorikken added @carmenbianca's commits to this PR. Sorry about not being able to push directly to the PR, that's a limitation of my employer's GitHub organization.

y0urself:
Is there any chance these arguments will get into a release anytime soon?

linozen:
Yes, we're planning a release in the next two weeks.

y0urself:
Thank you for the answer! :)

PR #624, opened 2022-11-12
README.md: Fix Package Manager Links

oxzi:
Changes:
- Add Alpine Linux URL.
- GNU Guix has bumped versions which are part of its URL.
- Fedora has changed its URL schema.
- The NixOS URL was especially to a now outdated version.

As an alternative, the growing list of different packages could be removed and only a link to repology.org next to a generic example - e.g. `apt-get install reuse` - could remain.


nicorikken (approved):
Great, thanks!

PR #625, opened 2022-11-15
Add support for file comments in spdx

alpianon:
For an open source project I'm working on ([Oniro](https://docs.oniroproject.org/en/latest/)), I added initial support for comments on single files, allowed both by dep5 and spdx specs:
- added 'comment' field to SpdxInfo
- added conversion of dep5 Files:Comment fields into SPDX FileComment fields
- TODO: it would be useful to be able also to *add* (human-readable) comments with `reuse addheader`

A typical (but not the only one) use case is to handle cases in which different portions of a file are subject to different licenses (I'm finding a lot of such cases with patch files in yocto layers). The official FAQ suggest to use comments to that purpose https://reuse.software/faq/#partial-license 

Specific tests are still missing. If there is consensus on this PR, I will add them (as proposed to @linozen )

### DIFF
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -77,7 +77,11 @@ _IGNORE_FILE_PATTERNS.extend(_IGNORE_SPDX_PATTERNS)
 #: The two iterables MUST be sets.
 SpdxInfo = NamedTuple(
     "SpdxInfo",
-    [("spdx_expressions", Set[Expression]), ("copyright_lines", Set[str])],
+    [
+      ("spdx_expressions", Set[Expression]),
+      ("copyright_lines", Set[str]),
+      ("comment", str)
+    ],
 )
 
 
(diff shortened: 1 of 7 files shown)


mxmehl:
Sorry for the late review. Please note that for marking portions of a file being under a different license, SPDX snippets are the way to go. They have been introduced in SPDX 2.3. We will also cover this in the next spec release. 

Here's the current status: https://github.com/fsfe/reuse-docs/blob/112d289667f2c4d315613a59ac467f9a6a3f45ef/spec.md?plain=1#L165

The FAQ item you referred to will also be updated: fsfe/reuse-docs#124

Apart from that, covering SPDX comments seems to be OK. But please note that there has been some work done on the SPDX output in a different PR which probably causes the conflict. I'm also not sure whether it conflicts with the big work going on in #654.

linozen:
Since #654 was merged, there now is a dataclass called [`ReuseInfo`](https://github.com/fsfe/reuse-tool/blob/main/src/reuse/__init__.py#L107-L122) to which SPDX comments need to be added. Is this still relevant, @alpianon? If yes, this PR would need to be rebased on `main` and specific tests should be added.

floriansnow:
No further activity after a ping + merge conflicts, so I'm closing this for now. Feel free to reopen if there is an update.

PR #632, opened 2022-12-04
New formats support

### LINKED ISSUE #575: reuse should detect .clang-format files
reuse should detect .clang-format files

```
$ reuse addheader --copyright="stenc authors" --license=GPL-2.0-or-later .clang-format 
usage: reuse addheader [-h] [--copyright COPYRIGHT] [--license LICENSE] [--year YEAR] [--style {applescript,aspx,bat,bibtex,c,css,f,ftl,handlebars,haskell,html,jinja,lisp,m4,ml,plantuml,python,rst,tex,vim}]
..reuse addheader: error: The following files do not have a recognised file extension. Please use --style, --force-dot-license or --skip-unrecognised:
.clang-format

```

### LINKED ISSUE #614: Comment style for proguard rules
Android projects typically contain `proguard-rules.pro` files that contains a configuration on optimizing code during the build process. More details are available [here](https://android-doc.github.io/tools/help/proguard.html) and [here](https://developer.android.com/studio/build/shrink-code). Android Studio creates the files with the default name `proguard-rules.pro` in each module directory.

Both `//` and `#` style comments should be supported, but I've only seen `#` in the wild and automatically generated template files.

My current proposal would be to add `proguard-rules.pro` files as a supported file to the comments with the `PythonCommentStyle` so the `#` is used.

### LINKED ISSUE #616: Support for .cjs and .mjs files
Please add support for the `.cjs` and `.mjs` file extensions. They are just alternatives to `.js` that specify what kind of module the file is (CommonJS, or ECMAScript). But they are both just JavaScript files, and should be handled in the same way (i.e., a C-style comment).

sgaist:
This merge request implements a couple of requested format support.

Each format is in it's own commit to keep things clearly separated.

Fixes #575
Fixes #614
Fixes #616

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@ The versions follow [semantic versioning](https://semver.org).
 
 - More file types are recognised:
   - Fennel (`.fnl`) (#638)
+  - CommonJS (`.cjs`) (#632)
+  - Qt .pro (`.pro`) (#632)
+- More files are recognised:
+  - Clang format (`.clang-format`) (#632)
 - Added loglevel argument to pytest and skip one test if loglevel is too high
   (#645).
 
--- a/src/reuse/comment.py
+++ b/src/reuse/comment.py
@@ -513,6 +513,7 @@ EXTENSION_COMMENT_STYLE_MAP = {
     ".bib": BibTexCommentStyle,
     ".c": CCommentStyle,
     ".cc": CCommentStyle,
+    ".cjs": CCommentStyle,
     ".cl": LispCommentStyle,
     ".clj": LispCommentStyle,
     ".cljc": LispCommentStyle,
--- a/src/reuse/comment.py
+++ b/src/reuse/comment.py
@@ -621,6 +621,7 @@ EXTENSION_COMMENT_STYLE_MAP = {
     ".pot": PythonCommentStyle,
     ".ppt": UncommentableCommentStyle,
     ".pptx": UncommentableCommentStyle,
+    ".pro": PythonCommentStyle,
     ".proto": CCommentStyle,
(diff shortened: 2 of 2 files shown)


nicorikken:
@sgaist Thanks a lot! Reading up on `.clang-format` I see this is supposed to be the entire filename, not the extension. Also the `_clang-format` filename is a valid alternative. I added it in https://github.com/fsfe/reuse-tool/issues/575 perhaps you can add that to that commit? Otherwise I think this looks good.

sgaist:
Sure thing !

Should we consider adding support for other tools configuration files like `.flake8`, `.bandit` and the likes ?

nicorikken:
If you have suggestions, please do. Please provide some examples or references in the commit or discussion so we can refer to it at a later stage.

sgaist:
Shouldn't I rather open an issue about that topic so we can discuss there the list of tools ?

sgaist:
In this case, should I remove the .clang-format commit from this PR and put it back when the tools file support discussion is done ?

sgaist:
Thanks @carmenbianca I somehow missed that one. Will be perfect for #633.

carmenbianca (approved):
LGTM

mxmehl (approved):

sgaist:
One thing I just realized, I have listed the specific issues per commit which can trigger quite a lot of noise on their respective page.

I think I should rewrite the message of these commits and only mention the issues in the pull request.

Agreed ?

sgaist:
I fixed the pylint issue and did the message change at the same time.

carmenbianca:
Added a change log. Will merge when tests succeed. Thank you @sgaist !

PR #636, opened 2022-12-11
tokenize license to fit 80-char row length

gpongelli:
Downloaded license file have longer row.

With this change the license file will look more familiar .

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ binaryornot = "^0.4.4"
 license-expression = ">=1.0"
 python-debian = "^0.1.38,!=0.1.45,!=0.1.46,!=0.1.47"
 setuptools = "*"
+nltk = "^3.7"
 
 [tool.poetry.dev-dependencies]
 Sphinx = ">=4.0.0"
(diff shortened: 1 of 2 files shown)


mxmehl:
I'm very hesitant towards mangling the license texts we downloaded from SPDX, even if I'd agree that a fixed-width makes them look more familiar.

But that's largely outside the scope of REUSE IMHO.

gpongelli:
Hi @mxmehl , the nltk library is used to truncate text without truncating words .

let me know if there’s interest on this PR and I’ll try to fix the pipeline, otherwise you can close the PR.

thanks.

carmenbianca:
The licences should be used as-is. If you want to wrap the licences, I think it's best to do this manually to make sure the wrapping software made no mistakes.

Even so, I recommend to simply make no changes to the file.

gpongelli:
Hi @carmenbianca , so do you think it's better to change the txt files that will be downloaded (like [this one for MIT](https://raw.githubusercontent.com/spdx/license-list-data/master/text/MIT.txt)) ?

carmenbianca:
That, or don't use `reuse download`, and instead copy the licence text from, say, https://choosealicense.com/licenses/mit/, where it's already wrapped.

Issue #642, opened 2023-01-11
Tests broken on github actions


carmenbianca:
The tests run against ubuntu:latest. ubuntu:latest does not ship with Python 3.6.

Fix this.

linozen:
Fixed as of #644

PR #644, opened 2023-01-11
Fix CI

linozen:
See related issue here:
https://github.com/actions/setup-python/issues/544


carmenbianca:
See #642

mxmehl (approved):

mxmehl:
Ah, just found an alternative solution. We could exclude 3.6 from the matrix, and include it explicitely with ubuntu-20.04. This way, we could run all other Python versions on ubuntu-22.04.

nicorikken:
That is what I was thinking too.

PR #645, opened 2023-01-11
Add loglevel argument to pytest. [convenience feature]

sisch:
While starting to work on reuse-tool I first ran the tests and noticed an overwhelming flood of DEBUG outputs.

This PR is my attempt to improve the user experince of running the tests frequently by allowing for a commandline option `--loglevel <LEVEL>`. This does break one test that relies on log output, thus this test is skipped, if the loglevel is too high.

Feel free to suggest a place to put some instructions on how to use it (if it isn't self-explanatory), as I wasn't sure where an example call would fit best.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ The versions follow [semantic versioning](https://semver.org).
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


carmenbianca (approved):
LGTM :+1: 

Thanks for the contribution @sisch !

PR #648, opened 2023-01-16
Rework documentation deployment

linozen:


.readthedocs.yaml
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: 2017 Free Software Foundation Europe e.V. <https://fsfe.org>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+version: 2
+
+build:
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.10"
```
carmenbianca, line 10: This should probably mirror our CI configuration, which uses 20.04 and Python 3.6, with a comment stating that these two should be identical.

.readthedocs.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+python:
+  install:
+    - requirements: docs/requirements.txt
```
carmenbianca, line 14: Was this file generated? Can you add a comment stating which command was used to generate the file?

linozen: For sure :) I was thinking of adding the command also to the `make docs` make instructions.

linozen: I did add it to `make docs` and created a separate `make docs-ci` to only run the sphinx build command when testing the generation process in the CI workflow.

PR #649, opened 2023-01-17
Bumping dependencies

linozen:
Should close #635


carmenbianca:
You may need to use poetry 1.0 to generate the lockfile.

linozen:
> You may need to use poetry 1.0 to generate the lockfile.

Yap, just trying to pin poetry to version 1.2.2 everywhere it's run so we have some consistency there.

carmenbianca:
> Yap, just trying to pin poetry to version 1.2.2 everywhere it's run so we have some consistency there.

I think this is incorrect, and I'm really surprised that the CI didn't break. We need Poetry at version <=1.1 for Python 3.6 support.

I don't really agree with this PR. Poetry shouldn't need to be pinned as explicitly (although doing this explicitly isn't necessarily bad).

Was it not sufficient to re-generate the lockfile using Poetry v1.1, which is our declared minimal version? Poetry should be forward-compatible.

.github/workflows/pythonpackage.yaml
```diff
@@ -25,12 +25,12 @@ jobs:
...
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install poetry
+          pipx install poetry==1.2.2
```
carmenbianca, line 33: Why doesn't this break on Python 3.6?

README.md
```diff
@@ -274,6 +274,13 @@ Next, you'll find the following commands handy:
 - `poetry run pylint src`
 - `make docs`
 
+Please note that we currently use version poetry version 1.2.2. You can make
+sure to install this version of poetry, e.g. by running:
+
+```bash
+pipx install poetry==1.2.2
+```
+
```
carmenbianca, line 283: I don't think this paragraph is necessary at all. Any version of Poetry >=1.1 should work, which is well-documented in pyproject.toml.

Even if we do document something here, it should say >=1.1, not a pinned version.

docker/Dockerfile-debian
```diff
@@ -18,7 +18,10 @@ RUN apt-get update \
     && apt-get install -y python3-dev python3-pip \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip install poetry
+RUN python3 -m pip install --user pipx
+RUN python3 -m pipx ensurepath
+# Make sure this is the same version as in the Alpine-based images
```
carmenbianca, line 23: I don't think this needs to be true. Once `python3-poetry` is available in Debian, we should just use that.

.github/workflows/pythonpackage.yaml
carmenbianca: Because, in switching to pipx, the tests now all run against Python 3.8.

https://github.com/fsfe/reuse-tool/actions/runs/3939665159/jobs/6739785792

This PR needs to be reverted.

linozen:
> Was it not sufficient to re-generate the lockfile using Poetry v1.1, which is our declared minimal version? Poetry should be forward-compatible.

Nope, I just can't get poetry<1.2 to resolve the dependencies - I'm always stuck on `Resolving dependencies...`. The problem was resolved when I used poetry>1.2.

PR #650, opened 2023-01-18
Bump certifi, adjust Poetry

carmenbianca:
Replaces #649, closes #635 

I ran `poetry update certifi` to get around https://github.com/python-poetry/poetry/issues/2094 on Poetry 1.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,8 +39,8 @@ The versions follow [semantic versioning](https://semver.org).
 
 ### Added
 
-- Added loglevel argument to pytest and skip one test if loglevel is too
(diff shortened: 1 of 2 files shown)


carmenbianca:
I tried to use the Poetry instructions from here, but it complained about finding no match for poetry-core>=1.1.0 when running `poetry install` -> https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

Let's just not touch Poetry for a while.

linozen:
Nice thanks @carmenbianca for taking a look. I agree, let's not touch poetry until we need to :)

PR #651, opened 2023-01-18
pipx install poetry in GitHub CI

carmenbianca:
#650 frustrated me, so I think I fixed it this time. Apparently `poetry-core` and `poetry` aren't necessarily in version sync (I think). When I installed poetry v1.1, I received poetry-core v1.0. When I installed poetry v1.0, I didn't get a poetry-core at all.

Best as I can gather, we need poetry v1.1 at a minimum, and also poetry-core v1.0 at a minimum. Later Poetry features are not supported/used, but later versions are backwards-compatible.

Instead of pinning Poetry to a specific version, I set a maximum, so that we get alerted in time in case Poetry breaks.

### DIFF
--- a/.github/workflows/pythonpackage.yaml
+++ b/.github/workflows/pythonpackage.yaml
@@ -24,13 +24,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Install poetry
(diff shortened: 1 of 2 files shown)


carmenbianca:
Of course this breaks on Windows.

PR #652, opened 2023-01-19
Do not abort the whole test job if one combination in the matrix fails

mxmehl:
We want to see the results of all combinations in the `test` job and not abort the whole job if only one fails.

This way it's easier to see in a PR which combinations are problematic, so we don't have to fix them one-by-one.


carmenbianca (approved):
LGTM

PR #653, opened 2023-01-21
Fix comment style for Scala

b-fein:
Scala uses the C comment style with `//` instead of `#` for both Scala 2 [1] and 3 [2].

[1] https://www.scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#whitespace-and-comments
[2] https://docs.scala-lang.org/scala3/book/scala-for-python-devs.html#comments


mxmehl (approved):
Thanks! That seems to have been an error, I cannot find a source that lists `#` as comment style in Scala.

carmenbianca:
Thank you @b-fein!

Issue #536, opened 2022-06-04
`reuse spdx` should exclude the tag from `FileCopyrightText`


Tachi107:
_(Mentioned in https://github.com/fsfe/reuse-tool/issues/59#issuecomment-1143691289, but opening as a separate issue because the request is not the same)_

The REUSE spec states that

> The `SPDX-FileCopyrightText` tag MUST be followed by a copyright notice.
>
> Instead of the `SPDX-FileCopyrightText` tag, the symbol `©`, or the word `Copyright` MAY be used, in which case a colon is not needed.

As `SPDX-FileCopyrightText`, `©`, and `Copyright` are _followed by_ the copyright notice, it means that they are not part of the notice itself, and shouldn't be part of the outputted `FileCopyrightText`.

In https://github.com/fsfe/reuse-tool/issues/59#issuecomment-519520939, @carmenbianca mentioned that _"It'd be fairly trivial to only take everything AFTER the copyright tag_".

Of course this is only my opinion, and I may be failing to see why including the various tags is important and not superfluous.

mxmehl:
Makes sense. To extend, Carmen was referring to everything after the matched copyright string (Copyright, C-symbol, SPDX etc), but that it's harder for just `SPDX-FileCopyrightText`. But it's certainly something that should be fixed in the long run.

mxmehl:
@silverhook could we please ask for your legal opinion here?

In today's call we were unsure what role `SPDX-FileCopyrightText: ` has. Is it equally *important* as, say, `Copyright © `?

Asked the other way round: If we removed `SPDX-FileCopyrightText` from the `reuse spdx` output in the `FileCopyrightText` fields, shouldn't we also remove e.g. `Copyright (c)` or `©`?

Our gut feeling, also stemmed from experiences with how sensible companies treat these copyright lines, was that the SPDX tag does not count to the actual copyright statement but just introduces it, unlike the "traditional" copyright statements.

silverhook:
@mxmehl , you mean from the theoretical copyright law PoV or from the practical compliance PoV?

From the copyright law PoV, as I [wrote in my blog post](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-the-sign), the only copyright notice formality that can be found in several copyright laws in the world is the actual `©` sign. Some (esp. English speaking) jurisdictions will also allow for the word “Copyright”, but others won’t. Due to the Berne convention though, everyone recognises the `©` sign.

[...]

PR #654, opened 2023-01-24
Adding `--json` to lint

### LINKED ISSUE #183: Add `--json` flag to lint

linozen:
This should hopefully close #183.

More tests need to be added and more feedback gathered. This is just a first stab at adding this functionality. Maybe we get it done pre-FOSDEM!

Many thanks!


mxmehl:
Thanks a bunch for implementing this! I'd also like to make @alpianon aware of this PR as he asked for the feature IIRC.

Then, let me post some remarks and questions:
1. My idea behind the `source` key in `licenses` and `copyrights` was that it should show where the information came from: the file directly, an accompanying .license file, or dep5? Currently, this always is `file.spdxfile.name`
2. I tested it with a repo containing *a lot* of issues, and it threw a traceback with `--json` added:

```
$ reuse lint --json
reuse.project - WARNING - Could not resolve SPDX License Identifier of LICENSES/InVaLiD.txt, resolving to InVaLiD. Make sure the license is in the license list found at <https://spdx.org/licenses/> or that it starts with 'LicenseRef-', and that it has a file extension.
reuse.report - ERROR - Could not read 'README.md'
PermissionError: [Errno 13] Permission denied: 'README.md'
  File "/home/max/.cache/pypoetry/virtualenvs/reuse-RJlAEOXQ-py3.10/bin/reuse", line 6, in <module>
    sys.exit(main())
  File "/home/max/Git/github/reuse-tool/src/reuse/_main.py", line 283, in main
    return parsed_args.func(parsed_args, project, out)
  File "/home/max/Git/github/reuse-tool/src/reuse/lint.py", line 331, in run
    lint(data, formatter=formatter, out=out)
  File "/home/max/Git/github/reuse-tool/src/reuse/lint.py", line 311, in lint
    out.write(formatter(data))
  File "/home/max/Git/github/reuse-tool/src/reuse/lint.py", line 295, in format_json
    return json.dumps(
  File "/usr/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/usr/lib/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 439, in _iterencode
    yield from _iterencode(o, _current_indent_level)
  File "/usr/lib/python3.10/json/encoder.py", line 429, in _iterencode
    yield from _iterencode_list(o, _current_indent_level)
  File "/usr/lib/python3.10/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 439, in _iterencode
    yield from _iterencode(o, _current_indent_level)
  File "/usr/lib/python3.10/json/encoder.py", line 436, in _iterencode
    raise ValueError("Circular reference detected")
ValueError: Circular reference detected
```

To reproduce:

```sh
git clone https://github.com/mxmehl/reuse-example.git test && cd test
git checkout so-many-errors
chmod 000 README.md   # also provoke a read error
reuse lint --json
```

The same error also occurs when linting the Linux kernel for example.

3. Then, copy from https://github.com/fsfe/reuse-tool/issues/183#issuecomment-1232962799: The `files` key also shows the `SPDX-FileCopyrightText` tag in the `value` key. As per #536 we consider to remove that from the `spdx` subcommand, so we may also want to do it here.

Otherwise, it looks good to me so far. It would be interesting to use it against a few concrete test cases.

linozen:
> Thanks a bunch for implementing this! I'd also like to make @alpianon aware of this PR as he asked for the feature IIRC.
> 
> Then, let me post some remarks and questions:
> 
>     1. My idea behind the `source` key in `licenses` and `copyrights` was that it should show where the information came from: the file directly, an accompanying .license file, or dep5? Currently, this always is `file.spdxfile.name`

This is now fixed by slightly adapting the `SpdxInfo` namedtuple. We decided that the move from a namedtuple to a dataclass is best handled in #669. We left a note in #669 about where `SpdxInfo` needs to be adapted.

>     2. I tested it with a repo containing _a lot_ of issues, and it threw a traceback with `--json` added:
> 
> 
> ```
> $ reuse lint --json
> reuse.project - WARNING - Could not resolve SPDX License Identifier of LICENSES/InVaLiD.txt, resolving to InVaLiD. Make sure the license is in the license list found at <https://spdx.org/licenses/> or that it starts with 'LicenseRef-', and that it has a file extension.
> reuse.report - ERROR - Could not read 'README.md'
> PermissionError: [Errno 13] Permission denied: 'README.md'
> Traceback (most recent call last):
>   File "/home/max/.cache/pypoetry/virtualenvs/reuse-RJlAEOXQ-py3.10/bin/reuse", line 6, in <module>
>     sys.exit(main())
>   File "/home/max/Git/github/reuse-tool/src/reuse/_main.py", line 283, in main
>     return parsed_args.func(parsed_args, project, out)
>   File "/home/max/Git/github/reuse-tool/src/reuse/lint.py", line 331, in run
>     lint(data, formatter=formatter, out=out)
>   File "/home/max/Git/github/reuse-tool/src/reuse/lint.py", line 311, in lint
>     out.write(formatter(data))
>   File "/home/max/Git/github/reuse-tool/src/reuse/lint.py", line 295, in format_json
>     return json.dumps(
>   File "/usr/lib/python3.10/json/__init__.py", line 238, in dumps
>     **kw).encode(obj)
>   File "/usr/lib/python3.10/json/encoder.py", line 201, in encode
>     chunks = list(chunks)
>   File "/usr/lib/python3.10/json/encoder.py", line 431, in _iterencode
>     yield from _iterencode_dict(o, _current_indent_level)
>   File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
>     yield from chunks
>   File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
>     yield from chunks
>   File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
>     yield from chunks
>   File "/usr/lib/python3.10/json/encoder.py", line 439, in _iterencode
>     yield from _iterencode(o, _current_indent_level)
>   File "/usr/lib/python3.10/json/encoder.py", line 429, in _iterencode
>     yield from _iterencode_list(o, _current_indent_level)
>   File "/usr/lib/python3.10/json/encoder.py", line 325, in _iterencode_list
>     yield from chunks
>   File "/usr/lib/python3.10/json/encoder.py", line 439, in _iterencode
>     yield from _iterencode(o, _current_indent_level)
>   File "/usr/lib/python3.10/json/encoder.py", line 436, in _iterencode
>     raise ValueError("Circular reference detected")
> ValueError: Circular reference detected
> ```
> 
> To reproduce:
> 
> ```shell
> git clone https://github.com/mxmehl/reuse-example.git test && cd test
> git checkout so-many-errors
> chmod 000 README.md   # also provoke a read error
> reuse lint --json
> ```
> 
> The same error also occurs when linting the Linux kernel for example.

This should now be fixed but using a custom JSON serializer. 

>     3. Then, copy from [Add `--json` flag to lint #183 (comment)](https://github.com/fsfe/reuse-tool/issues/183#issuecomment-1232962799): The `files` key also shows the `SPDX-FileCopyrightText` tag in the `value` key. As per [`reuse spdx` should exclude the tag from `FileCopyrightText` #536](https://github.com/fsfe/reuse-tool/issues/536) we consider to remove that from the `spdx` subcommand, so we may also want to do it here.

IMHO, this is probably best treated in a separate PR as it does not only relate to the JSON output of `lint`. What do you think?

mxmehl:
Thank you! I can a few first tests and noticed some issues. But first of all, I took the liberty to merge latest master in the branch.

> > 3. Then, copy from [Add `--json` flag to lint #183 (comment)](https://github.com/fsfe/reuse-tool/issues/183#issuecomment-1232962799): The `files` key also shows the `SPDX-FileCopyrightText` tag in the `value` key. As per [`reuse spdx` should exclude the tag from `FileCopyrightText` #536](https://github.com/fsfe/reuse-tool/issues/536) we consider to remove that from the `spdx` subcommand, so we may also want to do it here.
> 
> IMHO, this is probably best treated in a separate PR as it does not only relate to the JSON output of `lint`. What do you think?

Yes, that makes sense.

Now, to the issues I found:

1. I noticed that many outputs are appearing in a random order, e.g. the list of used licenses, both in the JSON and plain text output. Also, the files in the `files` key would be great to have sorted. It's not critical but a UX issue. Two runs of the program should result in the very same output.

2. I see some issues with the detection of where a file is described. Perhaps the core problem lies further down than what you touched in this PR, but it's clearly wrong. I made some test cases in my fork:

```sh
git clone https://github.com/mxmehl/reuse-example.git test && cd test
git checkout json-tests
```

## cat.jpg

With my edits, `cat.jpg` is described in both the `cat.jpg.license` as well as `.reuse/dep5`. The tool detects the different source, but does not mention them correctly here. Both the "Tester Testerson" copyright and the "Apache-2.0" license are in the dep5 file.

```json
    "img/cat.jpg": {
      "copyrights": [
        {
          "value": "SPDX-FileCopyrightText: 2017 Peter Janzen",
          "source": "img/cat.jpg.license"
        },
        {
          "value": "Tester Testerson",
          "source": "img/cat.jpg.license"
        }
      ],
      "licenses": [
        {
          "value": "CC-BY-4.0",
          "source": "img/cat.jpg.license"
        },
        {
          "value": "Apache-2.0",
          "source": "img/cat.jpg.license"
        }
      ]
    }
```

However, for `dog.jpg`, it correctly displays the origin as dep5, but for this image there is no `.license` file.

## .gitignore

That's even more complicated. The file itself contains only copyright info, the `.license` file a license and copyright, and dep5 a copyright and some invalid licensing info. This is the result:

```json
    ".gitignore": {
      "copyrights": [
        {
          "value": "SPDX-FileCopyrightText: 2019 Jane Doe <jane@example.com>",
          "source": ".gitignore.license"
        }
      ],
      "licenses": [
        {
          "value": "CC0-1.0",
          "source": ".gitignore.license"
        }
      ]
    }
```

As you see, the copyright in the file itself is ignored. I think this is intentional as we once decided that such a file can override all contents from the file itself to circumvent false-positive info. We may want to rethink that.

However, for some reason, the dep5 info is completely ignored here. I cannot explain why this is different from the `cat.jpg` case.

linozen:
Thanks for your feedback @mxmehl!

> * I noticed that many outputs are appearing in a random order, e.g. the list of used licenses, both in the JSON and plain text output. Also, the files in the `files` key would be great to have sorted. It's not critical but a UX issue. Two runs of the program should result in the very same output.

Yes, this was absolutely an issue. The last two commits should have fixed that for JSON and plaintext output respectively.

> * I see some issues with the detection of where a file is described. Perhaps the core problem lies further down than what you touched in this PR, but it's clearly wrong. I made some test cases in my fork.

Yes, I also expect the problem to have deeper roots then this PR. In a sense, we're now exposing REUSE internals (e.g. where a license/copyright information was gathered from) that were not exposed previously. Before JSON output, the tool only really needed to know _whether_ a file had license and copyright information attached to it and whether is info was valid. It didn't really matter where it came from. Now with the much more verbose JSON output we're facing the following issue:

It is unclear what to do when license and copyright information is described in two sources, i.e. a .license file and a dep5 file and which one takes precedence. This is related to [this docs issue about precedence](https://github.com/fsfe/reuse-docs/issues/70) and also [this issue about the undesired merging of copyright information from different sources](https://github.com/fsfe/reuse-docs/issues/115).

My approach to this would be:
1. Clearly define precedence in the the spec. Option 3 as outlined by @mxmehl in [the issue](https://github.com/fsfe/reuse-docs/issues/70#issue-669957038) seems to be most popular. It is this:

> file.license
> explicit in adjacent yaml (so file directly)
> explicit in faraway 
> file header
> glob in adjacent yaml (file covered e.g. as part of *.png)
> glob in faraway yaml

IMHO, we would need to include DEP5 here as well as we're not hard deprecating its use.

2. Implement precedence in this PR (keeping in mind that we don't have `REUSE.yaml` yet. The code where that happens is [here](https://github.com/fsfe/reuse-tool/blob/c36fd4657bbf43e945de63d3b5675907d72777ec/src/reuse/project.py#L136-L188) and it should not be too hard to (a) disable the merging behaviour described in [this issue](https://github.com/fsfe/reuse-docs/issues/115) and (b) implement the precedence logic to be implemented until we have `REUSE.yaml`.

What do you think?

src/reuse/lint.py
```diff
@@ -8,329 +8,260 @@
...
+        "--plain",
+        action="store_true",
+        help="Formats output as plain text",
+    )
+    mutex_group.add_argument(
+        "--format",
+        nargs="?",
+        choices=("json", "plain"),
+        help="Formats output using the chosen formatter",
+    )
```
carmenbianca, line 41: Why does this exist in tandem with `--json` and `--plain`?

I'd expect one way of defining the format, not two.

I don't have a strong preference between the two ways.

Also, `quiet` is missing here.

src/reuse/lint.py
```diff
@@ -8,329 +8,260 @@
...
+    """
+    output = ""
+    data = report.to_dict()
+
+    # If the project is not compliant:
+    if not data["summary"]["compliant"]:
+        # Bad licenses
+        bad_licenses = data["non_compliant"]["bad_licenses"]
+        if bad_licenses:
+            output += "# " + _("BAD LICENSES") + "\n\n"
```
carmenbianca, line 59: `output +=` is really bad for performance, especially for large strings (which this might well be), because a new string object is created every time (Python strings are immutable). Output should probably be a StringIO object (which is append-only mutable, afaik) that we convert to a string at the end.

src/reuse/lint.py
```diff
@@ -8,329 +8,260 @@
...
+        (
+            _("files with license information: {count} / {total}").format(
+                count=data["summary"]["files_with_licensing_info"],
+                total=data["summary"]["files_total"],
+            ),
+            "empty",
+        ),
+    ]
+
+    for key, value in summary_contents:
```
carmenbianca, line 197: `summary_contents` should be a dictionary if it has keys and values.

src/reuse/lint.py
```diff
@@ -8,329 +8,260 @@
...
+        raise TypeError(
+            f"Object of type {obj.__class__.__name__} is not JSON serializable"
         )
+
+    return json.dumps(
+        report.to_dict(),
+        indent=2,
+        # Serialize sets to lists
+        default=custom_serializer,
+        sort_keys=True,
```
carmenbianca, line 245: Are we sure we want to sort keys? This means that the `json_version` and `reuse_version` keys are no longer at the top (assuming that `json.dumps` respects Python's insertion-ordered dictionaries).

src/reuse/lint.py
```diff
@@ -8,329 +8,260 @@
...
+    :param report: Dictionary holding formatted ProjectReport data
+    :param formatter: Callable that formats the data dictionary
+    :param out: Where to output
+    """
 
+    out.write(formatter(report))
 
-def run(args, project: Project, out=sys.stdout):
+    data = report.to_dict()
+    result = data["summary"]["compliant"]
```
carmenbianca, line 260: I'm not certain this function should exist, or at least not exist in this shape. It does the following things:

- Print lint result using a formatter.
- Return whether the lint is compliant or not, which I'm not sure is related to the printing.
  - In order to do this, `to_dict()` is called on the report, which seems like a design error to me. There should be a Python function or property `is_compliant` on ProjectReport.

If we fix the remark in the third bullet point, then this becomes effectively:

```python
out.write(formatter(report))
return report.is_compliant
```

which doesn't really merit a function on its own, I think.

src/reuse/lint.py
```diff
@@ -339,6 +270,14 @@ def run(args, project: Project, out=sys.stdout):
     with contextlib.ExitStack() as stack:
         if args.quiet:
             out = stack.enter_context(open(os.devnull, "w", encoding="utf-8"))
```
carmenbianca, line 272: Instead of doing this, we could create a `format_quiet`.

carmenbianca (changes requested):
A lot of remarks. I like the bones of this PR. There's just a lot of nitpicking :sweat_smile:

src/reuse/report.py
```diff
@@ -74,30 +74,79 @@ def __init__(self, do_checksum: bool = True):
...
-                identifier: str(path)
-                for identifier, path in self.licenses_without_extension.items()
+        """Collects and formats data from report and returns it as a dictionary
+
+        :param report: ProjectReport object
+        :return: Dictionary containing data from the ProjectReport object
+        """
+        # Setup report data container
+        data = {
+            "json_version": "1.0",
```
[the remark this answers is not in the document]
carmenbianca: Renamed to `lint_version`.

src/reuse/project.py
```diff
@@ -137,30 +137,40 @@ def spdx_info_of(self, path: PathLike) -> SpdxInfo:
         """Return SPDX info of *path*.
 
         This function will return any SPDX information that it can find, both
-        from within the file and from the .reuse/dep5 file.
+        from within the file, the .license file and from the .reuse/dep5 file.
+
+        It also returns a single primary source path of the license/copyright
+        information, where 'primary' means '.license file' > 'header' > 'dep5'
```
carmenbianca, line 143: This doesn't really make sense to me. It should return the _correct_ source. If the returned `SpdxInfo` returns the aggregate of the file contents + the DEP5 file, then the source should be `'my_file'` for the lines extracted from the file, and `.reuse/dep5` for the lines extracted from DEP5.

This would change the return type of this function from `SpdxInfo` to `List[SpdxInfo]`. That's fine by me.

Furthermore, the source information should probably contain two grains of information:

- `source_type` which is an Enum à la `SourceType.FILE`, `SourceType.DOT_LICENSE`, `SourceType.DEP5` (and later `SourceType.REUSE_YAML`). This information is _programmatically_ useful.
- `source_path` which is a path to the exact file that contains the information. This information is _semantically_ useful.

If we want to implement that, we should probably change `SpdxInfo` into a dataclass first (and rename it to `ReuseInfo` while we're at it).

carmenbianca (changes requested):

src/reuse/report.py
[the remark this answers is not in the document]
carmenbianca: Opened issue #728 for the schema.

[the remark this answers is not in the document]
carmenbianca: `lint_version` was inspired by `lock-version` in `poetry.lock`: https://github.com/fsfe/reuse-tool/blob/bb3d76fe88298ad3ccc0c88320ef16ffaebb4aea/poetry.lock#L841

The rationale was that there could be other `*_version` fields if we output `--json` for other commands.

I did some research to see if there's a standardised way to do this by referencing the schema directly, like is often done in XML: there isn't. https://github.com/json-schema-org/json-schema-spec/issues/828

We could nevertheless just _do that_. Instead of `lint_version` (or other), use `{"$id": "https://reuse.software/link/to/1.0.0/lint_schema.json", ...}` (in the lint output). I chose `$id` because it most semantically matches the field used _inside of the schema_, see:

- https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.1 (still in draft)
- https://json-schema.org/learn/getting-started-step-by-step.html (User-friendly reading)

We could choose a different field, too. `$schema` and `schema` are also candidates.

In order to get all of that right, though, we have to commit to writing a json-schema in order to close this PR. Is that a commitment we want to do, or is that too much effort to get this rolling?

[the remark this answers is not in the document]
carmenbianca: I have just confirmed that vscode uses the `$schema` key to reference the schema, which then allows autocompletion by the editor.

src/reuse/lint.py
mxmehl: I personally wouldn't care about alnum sorting but about the fact that two runs of `reuse lint --json` return the very same output. Especially in larger projects I found it confusing in my tests that the order of keys and the list values aren't static.

If it were possible to sort all keys (and lists) but still have the version keys on top, that would be ideal. If not, I'd prefer the overall sorting to having the version on top.

src/reuse/report.py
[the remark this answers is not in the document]
carmenbianca: I'm not mixing issues :) Reiterating what you say to be on the page, the version we're discussing is specific to the output of `reuse lint --json`. If it changes (e.g., field gets added, renamed), that specific version gets bumped, and the package/spec versions change independently as needed.

The reason we may want a schema for the lint JSON output is so that programmers can use that as a sort of documentation for processing the output. The other reason for making a schema is that it would make the versioning of the output a lot easier to reason about; you can just `diff` the two different schema versions.

And in a scenario where we commit to creating schema files, we might as well put a link to the (versioned!) schema file in our output. This has the following benefits:

- We won't need to communicate the schema information out-of-band. `lint_version` refers to the version of a particular schema, but the link to the schema would have to be documented _elsewhere_. It unambiguously answers the question: what is this a version _of_?
- Increased visibility of the schema file. Programmers are lazy and don't read documentation. By putting a link to the schema file in the output itself, they may actually use it to inform their design choices.
- In a hypothetical future, including this key may enable _other_ programs to facilitate interactions with the file. That's what the vscode example was for; not to enable easier editing of the file, but to demonstrate that other programs can automatically detect the declared schema and use it to enhance certain operations. I was trying to detect what the exact field is that other programs are standardising on; it appears to be `$schema`.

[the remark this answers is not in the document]
mxmehl: OK, you convinced me of the advantages of schema files. So we'd include a `$schema` key. But how to version it then? Would the URL to the schema contain a version number? Or would that be a separate key?

In the case of the latter, I once again would prefer to not use the name `lint_version`, but instead `lint_schema_version` or so to clarify that this is not about the REUSE spec version.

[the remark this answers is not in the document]
carmenbianca: I don't think it's necessary to group the keys under `versions`. Otherwise I think we're in agreement. The power of consensus :D

src/reuse/lint.py
linozen: In the interest of getting this PR merged sooner rather than later, would it be OK for you two to create a new issue for this and then tackle it independently?

mxmehl: As said, I would consider unsorted keys on all levels as a bug and therefore important to be fixed in this PR. But yes, we could split off the features of having the `version` keys on the top of the output IMHO.

src/reuse/lint.py
```diff
@@ -8,329 +8,260 @@
 """
 
 import contextlib
+import json
 import os
 import sys
 from gettext import gettext as _
-from typing import Iterable
+from pathlib import PosixPath
```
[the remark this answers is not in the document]
linozen: No :)

src/reuse/lint.py
linozen: Done :)

src/reuse/lint.py
linozen: Just added a commit that should solve this.

src/reuse/lint.py
linozen: This is fixed in the newest commit.

src/reuse/project.py
carmenbianca: Ah, if we don't aggregate, then there's no need to refactor, you're right.

I still think it makes sense to rename `license_path` to `source_path` and to also encode `source_type`.

Also to rename `SpdxInfo` to `ReuseInfo`, but we can put that off and open another issue for that.

linozen: I'll resolve this as @carmenbianca already opened another issue for that.

src/reuse/report.py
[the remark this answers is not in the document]
linozen: I'll resolve this here now, as @carmenbianca already opened another issue for that.

src/reuse/lint.py
```diff
@@ -252,12 +252,24 @@ def custom_serializer(obj):
...
+            "lint_version": 0,
+            "reuse_spec_version": 1,
+            "reuse_tool_version": 2,
+        }
+        return (order.get(key, 3), key)
+
+    report_dict = report.to_dict_lint()
+    sorted_report_dict = dict(
+        sorted(report_dict.items(), key=lambda item: custom_sort_key(item[0]))
+    )
```
carmenbianca, line 254: Instead of sorting it _here_, can we sort it inside of `to_dict_lint`?

And then write a test for `to_dict_lint` to see whether it's sorted?

src/reuse/lint.py
linozen: This should be resolved now.

src/reuse/project.py
linozen: I encoded `source_type` not and refactored all the class and function names.

src/reuse/lint.py
linozen: Done :)

src/reuse/report.py
```diff
@@ -307,6 +357,7 @@ def __init__(self, name, spdx_id=None, chk_sum=None):
         self.chk_sum: str = chk_sum
         self.licenses_in_file: List[str] = []
```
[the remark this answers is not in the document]
linozen: I created a new issue for this: #753

floriansnow (approved):
LGTM!

src/reuse/__init__.py
```diff
@@ -89,12 +89,18 @@
...
-class SpdxInfo:
-    """Simple class holding SPDX information"""
+class ReuseInfo:
+    """Simple dataclass holding licensing and copyright information"""
 
     spdx_expressions: Set[Expression] = field(default_factory=set)
     copyright_lines: Set[str] = field(default_factory=set)
     contributor_lines: Set[str] = field(default_factory=set)
+    source_path: Optional[str] = None
+    source_type: Optional[str] = None
```
carmenbianca, line 99: This should be an Enum

linozen: It now is :)

PR #669, opened 2023-02-07
Implement support for SPDX-FileContributor

### LINKED ISSUE #615: feature request: add support for contributors to addHeader
This issue is a follow up of the discussion started in #612.

The idea is to add support for the `SPDX-FileContributor` field to the addHeader command so that one can set the copyright to an entity while still give credit to the authors of the file.

An example call I had in mind is:
```
reuse addheader --copyright="Example Labs ltd" --license=MIT --contributor="Charly Doe" --contributor="Jordan Doe" my_file.py
```

The result could look like the following:
```
# SPDX-FileCopyrightText: 2019 Example Labs ltd
#
# SPDX-FileContributor: Charly Doe
# SPDX-FileContributor: Jordan Doe
#
# SPDX-License-Identifier: MIT
```

_Originally posted by @sgaist in https://github.com/fsfe/reuse-tool/issues/612#issuecomment-1277791425_

sgaist:
This patch adds a new parameter that allows to list one or more contributors which will generate the corresponding list of fields in the header.

It refactors SpdxInfo into a dataclass so that it can be used with an optional list of contributors in a simple fashion.

Fixes #615


carmenbianca:
Hi @sgaist ! Thanks for this PR. It looks awesome. @linozen or me will review this quite soon, because we needed something just like this to implement another feature in #654.

We will also be dropping Python 3.6 support basically now-ish. If Python 3.6 is blocking you, please do ignore it.

sgaist:
Hi @carmenbianca ! You're welcome. There might be some work to be done on the test side as I only re-implemented two that were obvious good candidate but it might worth refactoring a bit more.

sgaist:
@sisch, ok, i'll wait for #654 to be merged and refactor this PR.

One related question: do you prefer that I merge main or rebase my work on it ?

sisch:
I'm afraid I need to pass this question to @linozen: do you know if rebase or merge is preferred?
Personally I think rebase in this specific case is better, but it's not my call, I'm not a maintainer ;)

mxmehl:
> @sisch, ok, i'll wait for #654 to be merged and refactor this PR.

Thank you for your patience!

> One related question: do you prefer that I merge main or rebase my work on it ?

In PRs, we usually prefer rebases. In this specific case I personally would also prefer a rebase unless a merge is much easier for you.

sgaist:
Rebase is perfect, that is my preferred way of working.

I'll keep an eye on #654 to keep things rolling.

carmenbianca:
I made some small changes and rebased this on top of main. I intend to merge this _before_ #654 because that PR would heavily benefit from the dataclass refactoring.

nicorikken:
Something to take note of, is that it is not possible to only provide the `--contributor` parameter:
```
$ reuse annotate --contributor "A Person" --recusvie myrepo/
 ... (usage info)
reuse annotate: error: option --copyright or --license is required
```

This can be a design choice, but I can imagine scenario's where a user would like to use a githook to add themselves as contributor, regardless of the copyright or license information. What do you think?

sgaist:
@nicorikken Good catch ! It's how I thought I had implemented it but did all the tests in the context of making it part of adding either of the other two fields.

src/reuse/__init__.py
```diff
@@ -82,13 +83,17 @@
...
-    [("spdx_expressions", Set[Expression]), ("copyright_lines", Set[str])],
-)
+
+@dataclass(frozen=True)
+class SpdxInfo:
+    """Simple class holding SPDX information"""
+
+    spdx_expressions: Set[Expression]
+    copyright_lines: Set[str]
+    contributor_lines: Set[str] = field(default_factory=set)
```
carmenbianca, line 93: I think we can do the default_factory for all of these

PR #673, opened 2023-02-08
Drop Python 3.6 support

### LINKED ISSUE #530: Drop support for Python 3.6
Python 3.6 has been end-of-lifed [for 5 months at time of writing](https://endoflife.date/python). Future development should not be beholden to Python 3.6.

carmenbianca:
Fixes #530 

This was surprisingly annoying. I thought Poetry was supposed to make life easier :weary: I'm stuck using Poetry v1.1 to maintain `poetry-core>=1.0.0` support, and it doesn't work quite as smoothly as later versions. Specifically, there was an endless loop for `sphinxcontrib.jquery`, which means I had to drop the RTD theme to preserve my patience for more important things.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,7 +21,12 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
-        os: [ubuntu-20.04, macos-latest, windows-latest]
+        os: [ubuntu-20.04]
+        include:
+          - python-version: "3.7"
+            os: macos-latest
+          - python-version: "3.7"
+            os: windows-latest
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


.github/workflows/test.yaml
```diff
@@ -20,11 +20,13 @@ jobs:
...
-        exclude:
-          - os: macos-latest
-            python-version: "3.6"
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        os: [ubuntu-20.04]
+        include:
+          - python-version: "3.7"
+            os: macos-latest
+          - python-version: "3.7"
+            os: windows-latest
```
nicorikken, line 29: Now we're not testing all Python versions on Windows and Mac, but I don't think this is an issue, right?

nicorikken (approved):
Great work, thanks! This will enable other improvements.

nicorikken:
@carmenbianca in the future we'll get to reap more of the tooling benefits :crossed_fingers:

.github/workflows/test.yaml
carmenbianca: Yes this is intentional. I just wanted to use fewer resources.

carmenbianca:
Thanks @nicorikken ! we can already re-enable pytest coverage after this PR :)

PR #674, opened 2023-02-09
Allow `build` module to interact with this Python project

### LINKED ISSUE #640: build.py causes FTBFS from alternative build tool
Hey, package maintainer from Arch Linux here. :wave: 

Just picking up our conversation from #600.

I can't seem to get reuse to build from source when using the [`build`](https://pypa-build.readthedocs.io/en/latest/) module to generate Python wheels.

<details><summary>Why do we use build? Well ...</summary>
Because there are ~5 (at last count) different ways to build a Python project that is PEP-517 compatible, hence us trying to only use the `build` module to build wheels rather than remembering the various CLI invocations for each individual tool. `build` is smart enough to hook into these other PEP-517 compatible builders and use their own build systems to generate a wheel/sdist archive.
</details>

The *only* hint I have of something breaking is running the following:

```
[root@chroot reuse]# python -m build --help
po/gl.po: warning: Charset "CHARSET" is not a portable encoding name.
                   Message conversion to user's charset might not work.
```
The command fails successfully, resulting in no wheel archive in `dist/`.

Any subsequent invocations of `python -m build --help` result in no output and no wheel archive in `dist/`:

```
[root@chroot reuse]# python -m build --help
[root@chroot reuse]#
```

<details><summary>For reference, this is the normal output:</summary>

```
$ python -m build --help
usage: python -m build [-h] [--version] [--sdist] [--wheel] [--outdir OUTDIR] [--skip-dependency-check] [--no-isolation] [--config-setting CONFIG_SETTING] [srcdir]

    A simple, correct PEP 517 build frontend.

    By default, a source distribution (sdist) is built from {srcdir}
    and a binary distribution (wheel) is built from the sdist.
    This is recommended as it will ensure the sdist can be used
    to build wheels.

    Pass -s/--sdist and/or -w/--wheel to build a specific distribution.
    If you do this, the default behavior will be disabled, and all
    artifacts will be built from {srcdir} (even if you combine
    -w/--wheel with -s/--sdist, the wheel will be built from {srcdir}).

positional arguments:
  srcdir                source directory (defaults to current directory)

options:
  -h, --help            show this help message and exit
  --version, -V         show program's version number and exit
  --sdist, -s           build a source distribution (disables the default behavior)
  --wheel, -w           build a wheel (disables the default behavior)
  --outdir OUTDIR, -o OUTDIR
                        output directory (defaults to {srcdir}/dist)
  --skip-dependency-check, -x
                        do not check that build dependencies are installed
  --no-isolation, -n    do not isolate the build in a virtual environment
  --config-setting CONFIG_SETTING, -C CONFIG_SETTING
                        pass options to the backend.  options which begin with a hyphen must be in the form of "--config-setting=--opt(=value)" or "-C--opt(=value)"
```

</details>

After flailing around, I've managed to find a horrible *horrible* hacky workaround that I most definitely am not proud of. Here be [dragons](https://github.com/archlinux/svntogit-community/tree/packages/reuse/trunk).

carmenbianca:
`build.py` was overriding the module. That's obviously not great.

Fixes #640 

Output:

```
$ python -m build
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (poetry-core>=1.0.0, setuptools)
* Getting dependencies for sdist...
* Building sdist...
* Building wheel from sdist
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (poetry-core>=1.0.0, setuptools)
* Getting dependencies for wheel...
* Building wheel...
Successfully built reuse-1.1.1.tar.gz and reuse-1.1.1-cp311-cp311-manylinux_2_36_x86_64.whl
```

Previously:

```
$ python -m build --help
$
```


carmenbianca:
Dropped in favour of #675

PR #685, opened 2023-02-16
Make translatable via weblate

### LINKED ISSUE #222: Implement translation through Weblate
Raised this in #217. Using this issue to track.

Link: https://hosted.weblate.org/projects/fsfe/reuse-tool/

carmenbianca:
- [x] Auto-generate po/reuse.pot after every push to `main` for weblate. Don't generate when nothing changes.
- [x] Documentation for contributors.
- [ ] ???
- [x] Check whether it actually works? I'm waiting for a weblate PR to this repo after I fixed something in weblate. #689

Fixes #222

### DIFF
--- a/.github/workflows/gettext.yaml
+++ b/.github/workflows/gettext.yaml
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2023 Carmen Bianca BAKKER <carmenbianca@fsfe.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+name: Update .pot file
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "src/reuse/**.py"
+
+jobs:
+  create-pot:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Install gettext
+        run: sudo apt-get install -y gettext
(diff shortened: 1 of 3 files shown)


.github/workflows/gettext.yaml
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: 2023 Carmen Bianca BAKKER <carmenbianca@fsfe.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+name: Update .pot file
+
+on:
+  push:
+    branches:
+      - main
```
nicorikken, line 10:
```suggestion
      - main
    paths:
      - 'src/reuse/**.py'
```

Besides the diff check after the fact, we can also ensure the Action is only run when relevant files have been edited.
GitHub Actions Documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

The current list of files considered for translation is here: https://github.com/fsfe/reuse-tool/blob/main/po/POTFILES.in

nicorikken (changes requested):
Small optimization suggestion. Otherwise great!

.github/workflows/gettext.yaml
carmenbianca: done

PR #691, opened 2023-02-18
Make _build.py compatible with msgfmt.py

carmenbianca:
If gettext is not installed, _build.py falls back to msgfmt.py, which is provided by upstream Python. However, msgfmt.py is more sensitive to the order of arguments.


mxmehl:
Probably not related, but in the Docker build we also have an error on msgfmt:

https://github.com/fsfe/reuse-tool/actions/runs/4211530610/jobs/7309939780#step:3:134

```
Building reuse (1.1.2)
  - Building sdist
  - Built reuse-1.1.2.tar.gz
  - Building wheel
/reuse-tool/_build.py:69: UserWarning: msgfmt is not installed. Translations will not be included.
  warn("msgfmt is not installed. Translations will not be included.")
  - Built reuse-1.1.2-cp310-cp310-musllinux_1_2_x86_64.whl
Removing intermediate container 0e171192ebba
 ---> 526f3ec9e4f8
```

Can this be addressed here as well, or is it a separate issue?

carmenbianca:
It's a separate issue. Thanks for noticing. Opened #693 to discuss that.

linozen:
LGTM :+1: Will merge now!

Issue #694, opened 2023-02-24
License text missmatch when downloading `GPL-2.0-only`


bernhardreiter:
There a section missing in the downloaded license for  `GPL-2.0-only`:

```bash
LANG=C reuse download GPL-2.0-only
Successfully downloaded LICENSES/GPL-2.0-only.txt

grep "does not permit" LICENSES/GPL-2.0-only.txt | wc
      0       0       0

curl --silent https://www.gnu.org/licenses/old-licenses/gpl-2.0-standalone.html | grep "does not permit"
This General Public License does not permit incorporating your program into

 reuse --version
reuse 1.1.2

```

The following section is missing from the download:

> This General Public License does not permit incorporating your program into proprietary programs. If your program is a subroutine library, you may consider it more useful to permit linking proprietary applications with the library. If this is what you want to do, use the GNU Lesser General Public License instead of this License.

mxmehl:
REUSE downloads the license texts from SPDX, specifically [license-list-data](https://github.com/spdx/license-list-data/tree/main/text). There, the paragraph you mention is marked as optional:

https://github.com/spdx/license-list-XML/blob/17632be2fcd4d4c2a9eda1202be2878baceb47b4/src/GPL-2.0-only.xml#L465-L473

Therefore, it does not appear in the final TXT file. If you think this is a mistake, I kindly ask you to send this request to SPDX directly, e.g. their legal mailing list.

We were asked a few times already to download the license texts from the authors directly but for technical, resource and maintenance reasons we rely on SPDX' data which usually is correct.

seabass-labrax:
Hello @bernhardreiter! SPDX Outreach Team co-chair here. SPDX considers that sentence to be 'optional text', so it's still considered to be `GPL-2.0-only` whether or not it includes that part. If you think that this sentence should be included in the text file, I would second @mxmehl's suggestion of writing to the spdx-legal@lists.spdx.org mailing list or the https://github.com/spdx/license-list-XML GitHub repository for consideration by the whole SPDX Legal Team :smiley:

bernhardreiter:
Hello @mxmehl, hello @seabass-labrax 
thanks for your responses!

Maybe there is room to improve the user experience for the reuse-tool and the SPDX licenses list,
as reuse lists the following URL as source for `GPL-2.0-only`:
  https://spdx.org/licenses/GPL-2.0-only.html

[...]

PR #695, opened 2023-02-26
chore: debug statement on license URL

nicorikken:
Adds a debugging statement to the license URL. Addresses the concern raised in https://github.com/fsfe/reuse-tool/issues/694

Signed-off-by: Nico Rikken <nico@nicorikken.eu>

### DIFF
--- a/src/reuse/download.py
+++ b/src/reuse/download.py
@@ -1,4 +1,5 @@
 # SPDX-FileCopyrightText: 2019 Free Software Foundation Europe e.V. <https://fsfe.org>
+# SPDX-FileCopyrightText: 2023 Nico Rikken <nico.rikken@fsfe.org>
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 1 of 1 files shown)


nicorikken:
Output when using debug:
```
reuse.download - DEBUG - license from 'https://raw.githubusercontent.com/spdx/license-list-data/master/text/gpl-2.0.txt
```

mxmehl (approved):
LGTM, thanks! Test failed, but not caused by this PR.

src/reuse/download.py
```diff
@@ -40,6 +41,7 @@ def download_license(spdx_identifier: str) -> str:
     """
     # This is fairly naive, but I can't see anything wrong with it.
     url = urljoin(_SPDX_REPOSITORY_BASE_URL, "".join((spdx_identifier, ".txt")))
+    _LOGGER.debug("license from '%s", url)
```
carmenbianca, line 44: missing operant verb 'downloading' :)

otherwise perfect

carmenbianca (dismissed):
will have a look at the failing test later. the test should probably be fixed before merging.

src/reuse/download.py
nicorikken: Check, fixed.

carmenbianca: I just noticed, the single quote isn't closed.

PR #699, opened 2023-03-10
Detect SPDX Snippets

### LINKED ISSUE #696: Fully support SPDX snippets
In REUSE spec version 3.1, we want to introduce support for [in-line SPDX snippets](https://spdx.github.io/spdx-spec/v2.3/file-tags/#h3-snippet-tags-format) (fsfe/reuse-docs#107). However, the tool does not support this yet.

Currently, the tool misses two features:

1. It does not detect the `SPDX-SnippetCopyrightText`.
2. The tool only reads the first 4 KB of a file (IIRC). As quite a few files tend to be larger than that, the tool might miss snippets further down the file.

Check out [this file](https://github.com/mxmehl/reuse-example/blob/snippet-testcase/src/main.c) that presents both problems in two snippets with [~4KB text in between](https://github.com/mxmehl/reuse-example/blob/c7e7a03438f8c043335fca8fbdf94cd3f25758b0/src/main.c#L22).

* The "global" copyright and license is detected.
* From the [first snippet by John (line 14-20)](https://github.com/mxmehl/reuse-example/blob/c7e7a03438f8c043335fca8fbdf94cd3f25758b0/src/main.c#L14-L20), only the MIT license is detected, no copyright.
* From the [second snippet by Hacker (line 24-30)](https://github.com/mxmehl/reuse-example/blob/c7e7a03438f8c043335fca8fbdf94cd3f25758b0/src/main.c#L24-L30), nothing is detected.

```
# SUMMARY

* Bad licenses:
* Deprecated licenses:
* Licenses without file extension:
* Missing licenses: MIT
* Unused licenses:
* Used licenses: CC-BY-4.0, CC0-1.0, GPL-3.0-or-later, MIT
* Read errors: 0
* Files with copyright information: 6 / 6
* Files with license information: 6 / 6
```

```
FileName: ./src/main.c
SPDXID: SPDXRef-63cc816bd7ec90aac6513f4273306341
FileChecksum: SHA1: a3ae3b0459e15ade0a6ecf992eef1ab222f508c8
LicenseConcluded: NOASSERTION
LicenseInfoInFile: GPL-3.0-or-later
LicenseInfoInFile: MIT
FileCopyrightText: <text>SPDX-FileCopyrightText: 2019 Jane Doe <jane@example.com></text>
```

## Possible solutions

Problem 1 shouldn't be hard to solve. As REUSE looks at the file level, we could treat the `SPDX-SnippetCopyrightText` tag the same as `SPDX-FileCopyrightText` so that it will appear in `reuse spdx` and, hopefully soon, `reuse lint --json` for instance.

Problem 2 is harder. We decided to only consider the first few kilobytes of each file as we assume that the interesting content is in the comment header which usually is not longer. But with snippets, interesting data might be at the very end of a very long file as well. So actually we would have to read the whole file.

However, there might be serious performance implications. I haven't run any tests with real repositories yet. We *could* make reading the whole file optional, either disabled or enabled by default.

mxmehl:
Fixes #696 

This PR shall consist of two steps:
1. Allow that the full file is analysed for license/copyright info if we have indicators that they contain a SPDX snippet
2. Actually understand the licensing/copyright info in a snippet (specifically the `SPDX-SnippetCopyrightText` tag)

On 1: Based on the results of the performance comparison, I implemented a logic that first scans the file for a string (as bytes) (`SPDX-SnippetBegin`), and resets the reading buffer. If a match is found, the `_HEADER_SIZE` limit (currently 4096 bytes) is unset so the full file is fed into `extract_spdx_info(decoded_text_from_binary())`.

Amazingly, the performance ramifications are minimal! On multiple runs against the Linux kernel, the negative effect is ~0.1s or ~1.3%). This is what I would consider an acceptable performance loss, especially in comparison with the 140% loss we'd have if we just removed the limit.

On 2: TODO


linozen:
LGTM :+1:

linozen (approved):

mxmehl:
> However, we _could_ check whether a snippet has licensing and copyright info. That may be a lot trickier as we would have to understand the boundaries of a snippet, and treat it isolatedly. I'm not sure whether it's worth to invest time into this now, or later. Interested in others' opinions.

Linus, Nico and I agreed that, while this would be interesting information, we don't plan to integrate this as of now. It could be a future improvement.

PR #701, opened 2023-03-13
Make translations available in Docker images

### LINKED ISSUE #693: gettext translations not included in Alpine Docker image
Opening this here for discussion.

gettext and/or `msgfmt.py` (from python-dev package? I'm not sure) are not installed in the Alpine Docker image, meaning that translations are not compiled/included.

mxmehl:
Fix #693 

gettext (msgfmt) was missing from the Docker images during build. This PR adds gettext during the setup phase. The actual Docker image size will not significantly be impacted as it's a multistage build.


linozen:
LGTM :+1:

PR #719, opened 2023-04-07
Fix gl.po

Gallaecio:
It [seems](https://groups.google.com/g/webpy/c/SxlC44z3VcY/m/Gjot1RioGfwJ) the reason I cannot use `reuse` in my machine is because one of these duplicate PO headers is wrong (`charset=CHARSET` should be `charset=UTF-8`).


Gallaecio:
I have just verified that installing the pre-commit hook from my branch indeed fixes the issue.

carmenbianca (approved):
LGTM, thank you very much!

PR #720, opened 2023-04-07
fix(docker): add git safe dir to data dir

janderssonse:
Fixes the requirement that: Newer versions of git needs the 'git config --global --add safe.directory /data' setting to work in a container.

Here is an example to verify both the problem and the fix: 

Problem:
1) .Build an reuse image from latest main code, official branch:

` docker build -f docker/Dockerfile-debian -t reusetest_default .`
 
2). Run the code in reuse compliant project, I used my script project (https://github.com/janderssonse/janderscripts)

` docker run --rm --volume $(pwd):/data reusetest_default lint`
 
 **Result**: Fails respecting .gitignore etc as the git config will fail and show warning '---no VCS'' etc.
 
 Fix:
3). Now try the same dockerfile but with the added fix in this pr:

` docker build -f docker/Dockerfile-debian -t reusetest_with_git_add_data_safe .`
 
4).  Run on same example repo as before (https://github.com/janderssonse/janderscripts)

` docker run --rm --volume $(pwd):/data reusetest_with_git_add_data_safe lint`
  
 ` Result`: success - git (vcs) accepted, .gitignore respected etc.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,6 +31,7 @@ COPY --from=build /opt/venv /opt/venv
 ENV VIRTUAL_ENV=/opt/venv
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
+RUN git config --global --add safe.directory /data
 WORKDIR /data
 
 ENTRYPOINT ["reuse"]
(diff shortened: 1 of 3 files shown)


linozen:
Many thanks @janderssonse for bringing this issue up and providing a fix right away. I tested this locally and all seems to work as expected again using your fix.

PR #724, opened 2023-04-07
Use importlib.metadata instead of pkg_resources

carmenbianca:
pkg_resources is deprecated. More information at <https://setuptools.pypa.io/en/latest/pkg_resources.html>.


linozen (commented):
LGTM :+1:

PR #726, opened 2023-04-09
Type checking with mypy

carmenbianca:
I'd been meaning to do this for a while. We have all these type annotations, but don't check whether they're valid.

### DIFF
--- a/src/reuse/_util.py
+++ b/src/reuse/_util.py
@@ -20,13 +20,14 @@ import shutil
 import subprocess
 import sys
 from argparse import ArgumentTypeError
+from collections import Counter
 from difflib import SequenceMatcher
 from gettext import gettext as _
 from hashlib import sha1
 from itertools import chain
 from os import PathLike
 from pathlib import Path
-from typing import BinaryIO, Dict, Iterator, List, Optional, Set
+from typing import IO, Any, BinaryIO, Dict, Iterator, List, Optional, Set, Union
 
 from boolean.boolean import Expression, ParseError
 from debian.copyright import Copyright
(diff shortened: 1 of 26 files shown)


carmenbianca:
Almost completed, except for `report.py` and `lint.py`. I'll wait on #654 to merge before doing that.

carmenbianca:
Not almost completed; I disallowed undefined and incomplete function defs, and will need to do some work on that.

src/reuse/_util.py
```diff
@@ -413,7 +426,12 @@ def _checksum(path: PathLike) -> str:
 class PathType:
     """Factory for creating Paths"""
 
-    def __init__(self, mode="r", force_file=False, force_directory=False):
+    def __init__(
+        self,
+        mode: str = "r",
```
carmenbianca, line 431: This type is probably incorrect

carmenbianca: Should be `Literal`, but only available in Python 3.8+.

nicorikken: You could add a comment for future reference:
```suggestion
        mode: str = "r", # change to Literal for python3.8+
```

nicorikken (approved):
My approval for the current changes, I reviewed all code changes. Thanks, great work!
Besides adding typing, I noticed a couple of improvements in logic, variables and handling like handling copyright statements. In the future perhaps specific types can be used for copyright statements and years as not to rely on `str`. I learned a couple of new constructs.

carmenbianca:
Rebased this on top of #759

src/reuse/_util.py
carmenbianca: I looked at this: there are too many possible permutations for me to bother with this. I'll just lazily keep it as-is.

PR #727, opened 2023-04-09
Don't use poetry>1.1 features

carmenbianca:
Related to #722 

Fixes problem introduced in #715, which had prior been fixed in #673, where it was introduced in #648 :sweat_smile:


nicorikken (approved):
Makes sense, good find and good to document 👍

PR #732, opened 2023-04-10
Remove `sphinxcontrib-apidoc` dependency

### LINKED ISSUE #731: `sphinxcontrib-apidoc` dependency is not maintained
https://github.com/sphinx-contrib/apidoc

Last commit 5 years ago. We should replace this.

carmenbianca:
Fixes #731


carmenbianca:
This will break the RTD build. Do not merge.

PR #743, opened 2023-05-16
add the moderatly commonly used extension .cxx as a c++ code

kjmeagher:


mxmehl (approved):
LTGM, thanks. Some proof for `.cxx` can be found [here](https://stackoverflow.com/questions/5171502/c-vs-cc-vs-cpp-vs-hpp-vs-h-vs-cxx).

A changelog entry isn't necessary as we don't add a new file type.

Issue #750, opened 2023-05-25
Add support for flutter pubspec.lock file "pubspec.lock" and flutter ".metadata" file


benoit-rolandeau-act:
## pubspec.lock

The `pubspec.lock` files are used by flutter to deal with dependencies management, see: https://dart.dev/tools/pub/glossary#lockfile. 
They have to be committed for application packages and not for library packages, see: https://dart.dev/guides/libraries/private-files#pubspeclock.
The file is written in yaml but generated by dart; therefore, we can't add manually information in it. That's why I think to use the class: UncommentableCommentStyle.

## .metadata

The `.metadata` files are generated by flutter when creating new packages. It contains the following comment at the file start:

> This file tracks properties of this Flutter project.
> Used by Flutter tool to assess capabilities and perform upgrades etc.
>
> This file should be version controlled and should not be manually edited.

It's written in yaml, but as explained in the comment, it can't be manually edited.

Moreover, I found that several other applications uses the `.metadata` extension, like: 
- Apple Metadata, and the file contains Apple Property List
- Acronis True Image Metadata Format

Therefore, I think the best class to use is: UncommentableCommentStyle

PR #751, opened 2023-05-25
Add support for flutter pubspec.lock file "pubspec.lock" and flutter ".metadata" file

benoit-rolandeau-act:
Add support for the pubspec.lock files used in Flutter to manage project dependencies.

Also support `.metadata` file which is used in Flutter to track project properties. The file can also be used by other libraries or in different context (but mostly as extension: `*.metadata` type instead of file).

Fixes https://github.com/fsfe/reuse-tool/issues/750


nicorikken:
Nice, simple and clean and with great explanation in the referred issue.

Issue #466, opened 2022-01-22
Replace python-debian dependency with statically provided code


mxmehl:
We want to get rid of python-debian as a dependency because it creates a number of problems on non-Debian/apt environments, and is generally containing a lot of code we don't need.

Instead, let's export the relevant code and ship it with reuse-tool directly.

Would fix #425 and #427, and probably make fixing issues like #454, #283, #191 and #163 much easier.

llimeht:
Writing on behalf of the python-debian maintainer team:

We want python-debian to be portable and widely usable. We definitely don't want people to feel the need to fork it, particularly not for simple things like portability of Python code. 

We've got a stack of portability fixes ready for the next release, with all tests now passing under Debian, Alpine, a stripped-down-Debian, and Windows. Are there other problematic hosts we should test against? We're happy to hear from you if you find other bugs or if you have other enhancements for python-debian.

Tachi107:
As this hasn't been mentioned before, I think that [nexB/debian-inspector](https://github.com/nexB/debian-inspector) might be an interesting alternative in the case that python-debian still doesn't suit your needs. (Cc: @pombredanne)

floriansnow:
Since dep5 is deprecated in REUSE, I don't think this issue is necessary any longer.

PR #752, opened 2023-05-31
feat: debian-inspector

nicorikken:
Use nexB/debian-inspector instead of python-debian to address #466 and hopefully improve on other dep5-parsing related issues.

_This PR is very much a work in progress_

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -27,7 +27,7 @@ wrapt = [
 name = "attrs"
 version = "22.2.0"
 description = "Classes Without Boilerplate"
-category = "dev"
+category = "main"
 optional = false
 python-versions = ">=3.6"
 
@@ -181,7 +181,7 @@ test = ["flake8 (==3.7.8)", "hypothesis (==3.55.3)"]
 
 [[package]]
 name = "coverage"
-version = "7.2.3"
+version = "7.2.6"
 description = "Code coverage measurement for Python"
 category = "dev"
 optional = false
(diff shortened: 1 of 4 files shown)


nicorikken:
Example structure of parsed dep5 file:
`
DebianCopyright(paragraphs=[CopyrightFilesParagraph(line_numbers_by_field={'files': (1, 1), 'copyright': (2, 2), 'copyright_1': (3, 3), 'license': (4, 4)}, files=AnyWhiteSpaceSeparatedField(values=['doc/*']), copyright=CopyrightField(statements=[CopyrightStatementField(holder='Jane Doe', year_range='2017')]), license=LicenseField(name='CC0-1.0', text=[]), comment=FormattedTextField(text=None), extra_data={'copyright_1': '2017 Jane Dox'}), CopyrightFilesParagraph(line_numbers_by_field={'files': (6, 6), 'copyright': (7, 7), 'copyright_1': (8, 8), 'license': (9, 9)}, files=AnyWhiteSpaceSeparatedField(values=['src/*']), copyright=CopyrightField(statements=[CopyrightStatementField(holder='Jane Doe', year_range='2017')]), license=LicenseField(name='CC0-1.0', text=[]), comment=FormattedTextField(text=None), extra_data={'copyright_1': '2017 Jane Dox'}), CopyrightHeaderParagraph(line_numbers_by_field={'format': (11, 11), 'upstream_name': (12, 12), 'upstream_contact': (13, 13), 'source': (14, 14)}, format=SingleLineField(value='https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/'), upstream_name=SingleLineField(value='Some project'), upstream_contact=LineSeparatedField(values=['Jane Doe']), source=FormattedTextField(text='https://example.com/'), disclaimer=FormattedTextField(text=None), copyright=CopyrightField(statements=[]), license=LicenseField(name='', text=None), comment=FormattedTextField(text=None), files_excluded=AnyWhiteSpaceSeparatedField(values=[]), extra_data={})])
`

nicorikken:
The debian-inspector doesn't have a convenient function to check is a path is covered and select the relevant dep5 statement, so it would have to be matched ourselves. Similar to https://github.com/nexB/debian-inspector/blob/240e24a559c2d91c0f2d6b79a7c7f4891e3dc582/src/debian_inspector/coverage.py#L35

floriansnow:
>  `debian-inspector` isn't packaged there yet, and it's unlikely that it will be in Debian Stable for the next 2 years.

I think the idea would be to supply it statically. But looking into this a bit, I also wonder if we could stick with the original solution because of this comment: https://github.com/fsfe/reuse-tool/issues/466#issuecomment-1284708611 I don't have a good overview of the exact problems in the past (especially because I use Debian), but perhaps that would be worthwhile looking into.

carmenbianca:
I think this is fixed by #841

PR #755, opened 2023-06-03
Add a set of Qt file types (pri, qrc, qss, ui)

anthony-loiseau-act:
Add few more Qt-related file extensions to known ones:

**.pri**: They are very closed to already supported .pro files.
      pri can also refers to binary files in Window app context, but this is not an issue to mark them as python-like files because reuse enforces .licence file for binary files.

**.qrc**: Those are Qt XML files containing indexes to resource files.
      Added as Uncommentable because Qt Creator IDE rewrites the file when saving, loosing copyrights each time therefore loosing history.

**.qss**: Qt Style Sheet, based on CSS file format.

**.ui**: Qt user interface definition file, XML.
     Added as Uncommentable because ui files are overwritten by Qt Creator upon saving which looses copyright history.


carmenbianca:
I think it's OK to mark .pri with a comment style. I may have to double-check this, but I believe that the program checks _anyway_ whether a given file is binary.

anthony-loiseau-act:
Thanks @carmenbianca for this note, you are right. Reuse header.py is_uncommentable function unconditionally returns true for files detected as binary.
I tested with a legitimate text.pri file and a fake binary.pri file, reuse properly injected licence in text.pri and created binary.pri.license 👍 

Therefore I updated my PR to declare Qt pri files as python-like files (as for .pro files)

linozen:
Thanks for the update @anthony-loiseau-act. Will merge now!

PR #759, opened 2023-06-10
Some followup work on the `--json` PR

carmenbianca:
Some refactoring work on top of the work done in #654.


src/reuse/header.py
```diff
@@ -124,30 +124,30 @@ def _create_new_header(
...
+    """Create a header containing *reuse_info*. *header* is an optional argument
+    containing a header which should be modified to include *reuse_info*. If
     *header* is not given, a brand new header is created.
 
     *template*, *template_is_commented*, and *style* determine what the header
     will look like, and whether it will be commented or not.
 
     :raises CommentCreateError: if a comment could not be created.
-    :raises MissingSpdxInfo: if the generated comment is missing SPDX
+    :raises MissingReusInfo: if the generated comment is missing SPDX
```
carmenbianca, line 150: typo

src/reuse/__init__.py
```diff
@@ -114,13 +109,56 @@ class ReuseInfo:
...
+        >>> one = ReuseInfo(copyright_lines={"Jane Doe"}, source_path="foo.py")
+        >>> two = ReuseInfo(copyright_lines={"John Doe"}, source_path="bar.py")
+        >>> result = one.union(two)
+        >>> print(sorted(result.copyright_lines))
+        ['Jane Doe', 'John Doe']
+        >>> print(result.source_path)
+        foo.py
+        """
+        new_kwargs = {}
+        for key, attr_val in self.__dict__.items():
```
nicorikken, line 145: Just checking: there is no need to check the keys of the other dictionary as all keys are added on initiation and so the structure is the same.

nicorikken (approved):
Lots of great improvements under the hood. Nice!

linozen (approved):
All looks great to me. Thanks @carmenbianca

src/reuse/__init__.py
carmenbianca: Correct, provided you have two `ReuseInfo` objects.

I considered the cases `ReuseInfo.copy(SuperReuseInfo)` and `SuperReuseInfo.copy(ReuseInfo)`, but didn't write tests for them.

In the first case, the extra fields in `SuperReuseInfo` are not copied over into the new object, because they're not in `ReuseInfo`'s __dict__.

In the second case, the extra fields in `SuperReuseInfo` _cannot_ be overwritten because they do not exist in `ReuseInfo`.

PR #760, opened 2023-06-15
Explicitly state deprecation instead of trying to suppress it

### LINKED ISSUE #739: Confusing CLI help sub-command output
Running from `$ pip install git+https://github.com/fsfe/reuse-tool.git` it looks like there's a bug in the `argparse` (or `click` or whatever you're using) CLI sub-command configuration for the, I'm guessing, deprecated `addheader` sub-command:

```
$ reuse --help
usage: reuse [-h] [--debug] [--include-submodules] [--include-meson-subprojects]
             [--no-multiprocessing] [--root PATH] [--version]
             {annotate,addheader,download,init,lint,spdx,supported-licenses,supported-licences}
             ...

reuse is a tool for compliance with the REUSE recommendations. See
<https://reuse.software/> for more information, and
<https://reuse.readthedocs.io/> for the online documentation.

This version of reuse is compatible with version 3.0 of the REUSE
Specification.

Support the FSFE's work:

  Donations are critical to our strength and autonomy. They enable us to
  continue working for Free Software wherever necessary. Please consider
  making a donation at <https://fsfe.org/donate/>.

options:
  -h, --help            show this help message and exit
  --debug               enable debug statements
  --include-submodules  do not skip over Git submodules
  --include-meson-subprojects
                        do not skip over Meson subprojects
  --no-multiprocessing  do not use multiprocessing
  --root PATH           define root of project
  --version             show program's version number and exit

subcommands:
  {annotate,addheader,download,init,lint,spdx,supported-licenses,supported-licences}
    annotate            add copyright and licensing into the header of files
    addheader           ==SUPPRESS==
    download            download a license and place it in the LICENSES/ directory
    init                initialize REUSE project
    lint                list all non-compliant files
    spdx                print the project's bill of materials in SPDX format
    supported-licenses (supported-licences)
                        list all supported SPDX licenses
```

floriansnow:
Apparently, argparse does not support suppression of help for subcommands and this will likely not change:
https://github.com/python/cpython/issues/67037
The simple, even if not quite as pretty solution is to just make the deprecation explicit instead of trying to suppress it.

Closes #739.

### DIFF
--- a/src/reuse/_main.py
+++ b/src/reuse/_main.py
@@ -150,7 +150,7 @@ def parser() -> argparse.ArgumentParser:
         "addheader",
         header.add_arguments,
         header.run,
-        help=argparse.SUPPRESS,
+        help=_("deprecated in favor of annotate"),
     )
 
     add_command(


carmenbianca:
It can be done, but none of the solutions are especially pretty: https://stackoverflow.com/questions/21692395/hiding-selected-subcommands-using-argparse

floriansnow:
> It can be done, but none of the solutions are especially pretty: https://stackoverflow.com/questions/21692395/hiding-selected-subcommands-using-argparse

I know. I looked at those threads as well and decided against just stating that this option is deprecated. But if you prefer, we can use a different solution.

carmenbianca (approved):
@floriansnow No, I agree with your reasoning :) LGTM. You should just wrap the string in `_("string here")` for i18n.

floriansnow:
Good catch! The things you miss after not being able to code for a year. 🙈

PR #761, opened 2023-06-16
Default values for `reuse annotate`, enabling a pre-commit hook

### LINKED ISSUE #68: Add configuration for addheader and provide --defaults for addheader

### LINKED ISSUE #534: Script documentation: run `addheader` on new files via pre-commit
@carmenbianca desires a feature that somehow automatically takes care of adding REUSE headers to new files, ideally via a pre-commit hook. Ideally, this would also be configurable per repo, as mentioned in #68.

The tool maintainers agreed that this should not be included in the tool itself, because there are a number of difficulties:

* Users of this feature may overwrite existing headers (see the discussion in #338)
* They may also introduce false information that does not match the actual copyright/licensing of the project or file that's added. Example: a users pastes a file from a different project under different copyright/license that does not carry REUSE information yet. Such an automatic hook would out false info on this file!

Instead, we should document how developers could do this in the [scripts documentation](https://reuse.readthedocs.io/en/stable/scripts.html). while making the aware of the potential issues.

carmenbianca:
Fixes #68
Fixes #534 

WIP


src/reuse/_config.py
```diff
@@ -0,0 +1,103 @@
...
+    # TODO: We could probably smartly cache the results somehow.
+    def annotations_for_path(self, path: PathLike) -> AnnotateOptions:
+        """TODO: Document the precise behaviour."""
+        path = Path(path)
+        result = self.global_annotate_options
+        # This assumes that the override options are ordered by reverse
+        # precedence.
+        for o_path, options in self.override_annotate_options.items():
+            o_path = Path(o_path).expanduser()
+            if path.is_relative_to(o_path):
```
carmenbianca, line 93: Apparently this method doesn't exist in Python ≤3.8.

Huy-Ngo: I think it can be replaced with `relative_to`, which exists in 3.8 though. From the 3.9 source:

```python
    def is_relative_to(self, *other):
        """Return True if the path is relative to another path or False.
        """
        try:
            self.relative_to(*other)
            return True
        except ValueError:
            return False
```

In newer versions both functions are based on `with_segments` which doesn't exist in 3.8 or 3.9 yet.

floriansnow:
Is there still interest in these changes?

Issue #202, opened 2020-04-14
Ignore relative symlinks contained inside the repo


amerlyq:
External file license is useful for binary files (because there is no way to add header to them) and auto-generated config files which strip-out any added comments (like linux-kernel "make savedefconfig").
In the latter case it's often useful to have relative symlink to default config to achive explicit transparecy inside project.
But it requires symlinking both file and its license for `reuse lint` to ignore that file.
Example:
```bash
mkdir relsymlink
cd relsymlink
git init
touch blob
reuse addheader -c user -l GPL-3.0-only --explicit-license -- ./blob
reuse download GPL-3.0-only
reuse lint  #1
ln -srT blob symlink
reuse lint  #2
ln -srT blob.license symlink.license
reuse lint  #3
```
In the case `#2` above license of `symlink` file won't be found.
And for `#3` to work you must symlink license to every place you symlinked the config (yes, it's rare but I already had necessity to do it thrice in last 6 months).

Possible solutions (limited only to relative symlinks fully contained inside the repo):
1) dereference symlink and require to have .license file only beside the original file.
2) (better) ignore all relative symlins (because they aren't subject to licensing anyway, and their original file will be checked for license anyway.

mxmehl:
Since symlinks could be rather understood as text files just containing a file path, I guess it could be ignored by REUSE. On the other hand, REUSE mandates to provide copyright/licensing information for *every* file, including the most irrelevant ones.

Not sure how symlinks would interact with this principle.

carmenbianca:
Ignoring symlinks is probably fine.

amerlyq:
I'm sure copyright and licensing laws can't be applied to symlinks at all.
There is no intellectual property to defend. So I don't see the point of
being pedantic to the bones on this matter.

Especially when enforcing it has the significant limiting drawback of
preventing people using symlinks inside their repo at all and creating ugly
workarounds on their place only to be pedantically compliant at all costs.

silverhook:
I would agree that symlinks, similarly as zero-size files (see https://github.com/fsfe/reuse-tool/issues/208#issuecomment-630220745) are safe to be ignored by the tool and spec.

PR #764, opened 2023-06-19
Fix symlink handling

### LINKED ISSUE #627: Symbolic links (git-annex files) are ignored
[git-annex](https://git-annex.branchable.com) is a tool to manage big files in Git repositories. In science it is used by the [Datalad](https://datalad.org) community to manage dataset. Git-annex works by managing **symbolic links** in the Git work tree which point to the actual file conten in `.git/annex/objects`. Now, `reuse` seems to ignore such symbolic links.

## Steps to reproduce
First issue is that `reuse lint` doesn’t complain about annexed files not having a license.

```sh
git init test
cd test
git annex init
echo "hello" > my_file
git annex add my_file # The file becomes a symbolic link.
reuse lint # Good result despite missing license
```

This becomes worse if I do assign a license, but then `reuse lint` *fails* because of an “unused” license. This leads to failed CI pipelines. So to continue the above:

```sh
reuse addheader -l'CC0-1.0' -c'author' --force-dot-license my_file
reuse download "CC0-1.0"
reuse lint # fails despite CC0-1.0 being used
```

Reuse version: 1.0.0  
git-annex version: 10.20221103

## Desired behavior
`reuse` follows symlinks, at least if they are annexed files, which means that the symlink points to something in `.git/annex/objects`.

---

I see that issue #202 discussed the topic of symlinks. I suggest to revisit the issue for git-annex.

matrss:
This PR implements the heuristics for handling symlinks I've described in more detail in https://github.com/fsfe/reuse-tool/issues/627#issuecomment-1418976389 and https://github.com/fsfe/reuse-tool/issues/627#issuecomment-1422594763.

Basically, this is based on an interpretation of the REUSE specification that:
1. a symlink pointing to a covered file is considered to be the same file as the covered file and can therefore be ignored.
2. a symlink pointing to a file that is not a covered file is itself considered to be a covered file and should not be ignored, unless the symlink is ignored by other means.

This fixes #627 while not breaking #202 for the described situation, since it would fall into category 1 above.


matrss:
@wtraylor Feel free to try this out on a real-world project if you have one available. This was less straight-forward to implement than I thought it would be and there might still be some edge-cases I didn't consider. I tried to implement the heuristics I thought to be intuitive, but that might also differ from yours. So some testing would be great :)

carmenbianca:
Hi @matrss 

This PR looks qualitatively really good. We'll need some time to review it, though.

It will also need a change to the specification. Although this version of the spec is not yet live, [the current development version says that symlinks are ignored](https://github.com/fsfe/reuse-docs/blob/0913b0a83b36c161966be1c5e70c81bdadfb8a69/spec.md#definitions).

wtraylor:
Many thanks for implementing this feature, @matrss !
For the main use case with `git-annex` it works great: `reuse lint` correctly complains about annexed files (i.e., symlinks pointing to content in `.git`) not having license information.

```
git init repo && cd repo
git annex init
cp ~/bigfile.dat ./
git annex add bigfile.dat # creates a symlink in place of bigfile.dat
reuse lint  # fails correctly
reuse annotate --force-dot-license -l CC0-1.0 -c Author bigfile.dat
reuse download --all
reuse lint # passes correctly
```

The implementation appears to assume that all symlinks require dot-license files. For example:

```
git init repo && cd repo
git annex init
echo Hello World > README.md
reuse annotate -l CC0-1.0 -c Me README.md
reuse download --all
reuse lint # success
rm -r LICENSES
git annex add README.md
git commit -m "Add annexed README"
reuse lint # fails: no licensing info for README.md
reuse annotate -l CC0-1.0 -c Me README.md # fails: can’t write to README.md
```

I guess it makes sense to require dot-license files for symlinks because the symlink targets may not be available (i.e., symlink is broken, for example if annexed content is not in the working copy). However, I suggest to make this more transparent to the user—for example with these messages:

1. `reuse annotate` on a symlink file: “blablabla is a symbolic link. Please use --force-dot-license.”
2. `reuse lint` with a symlink file that doesn’t have a dot-license: “The following symbolic links have no copyright and licensing information in separate dot-license files:”
3. `reuse lint` with a symlinked dot-license file: “WARNING: Can’t parse blablabla.license because it’s a symbolic link. Dot-license files should be regular files.”

Point 3 stems from my experience that it can easily happen to annex a file by accident. Then you might end up with an annexed `.license` file that REUSE doesn’t parse.

Finally, there is the question of how to deal with **symlink folders.** This does not fall into the use case of `git-annex` (which only symlinks files). Here’s an example:

```
mkdir folder
echo Hello World > folder/foo
git init repo && cd repo
ln -s ../folder symlinked_folder
reuse lint # no complaints
```

I think this behavior is good. It doesn’t make sense for `reuse` to check files in symlinked folders with content outside of the repository. However, I wonder if it would make sense to make this transparent to the user. For example, `reuse lint` good give a hint that it ignored `symlinked_folder` because it’s a symbolic link.

matrss:
I would like to get back to this, but I am seeing that the reuse spec that unconditionally ignores all symlinks is now live, and the above linked repository is archived. The readme in the reuse-website repo also links to the archived reuse-docs repository for the specification. Where do I have to suggest a change to the spec? Should I send a PR to reuse-website changing the spec-3.3.md directly?

PR #778, opened 2023-06-21
Revert the mandatory order of precedence; display PendingDeprecationWarning

carmenbianca:
See rationale in #779 for reverting this behaviour.

### DIFF
--- a/src/reuse/project.py
+++ b/src/reuse/project.py
@@ -10,6 +10,7 @@ import contextlib
 import glob
 import logging
 import os
+import warnings
 from gettext import gettext as _
 from pathlib import Path
 from typing import Dict, Iterator, Optional, Union, cast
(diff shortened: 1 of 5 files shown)


src/reuse/project.py
```diff
@@ -217,26 +219,36 @@ def reuse_info_of(self, path: StrPath) -> ReuseInfo:
...
-                    " '{dep5_path}'. The information in the DEP5 file has been"
-                    " overridden. Please ensure that this is correct."
+                    "Copyright and licensing information for"
+                    " '{original_path}' has been found in both '{path}' and"
+                    " in the DEP5 file located at '{dep5_path}'. The"
+                    " information for these two sources has been"
+                    " aggregated. In the future this behaviour will change,"
+                    " and you will need to explicitly enable aggregation."
+                    " See"
+                    " <https://github.com/fsfe/reuse-docs/issues/70>. You"
```
carmenbianca, line 232: Is this the correct URL?

carmenbianca: Created it here: #779

linozen (approved):
These changes are just what we need right now. It uses all the existing logic to revert to aggregation (for now) and can be merged as is IMHO. 

Thanks for all your effort @carmenbianca

PR #783, opened 2023-06-22
Add workflow to lint 3rd party repositories

### LINKED ISSUE #782: Check 3rd party repos before release or as regular CI job
At least as part of the release checklist, we should run the upcoming version of the linter against third-party repositories for which we know that they are
* rather complex (like curl)
* may use some "tricks" (dep5) (SAP repos come to mind)
* are reliable REUSE adopters and keen to keep this status
* are somewhat popular

This could be put into the [checklist](https://github.com/fsfe/reuse-tool/blob/main/CONTRIBUTING.md#release-checklist) or as part of the regular checks in CI. There, we could pin these repos to specific commits to avoid false-negatives, but perhaps that's even not necessary as they run the linter themselves before adding commits.

mxmehl:
Fixes #782


.github/workflows/3rd_party_lint.yaml
```diff
@@ -0,0 +1,51 @@
...
+          pip install poetry
+          poetry install --no-interaction --no-dev
+      # Clone and lint repositories
+      - name: Clone ${{ matrix.repo }}
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ matrix.repo }}
+          path: 3rd/${{ matrix.repo }}
+      - name: Lint ${{ matrix.repo }}
+        run: poetry run reuse --root 3rd/${{ matrix.repo }} lint
```
carmenbianca, line 51: I think it'd be nice if we instead called a script that does this, so that we can call the script locally during testing.

.github/workflows/3rd_party_lint.yaml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  third-party-lint:
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.ignore-failure }}
+    strategy:
+      # do not abort the whole test job if one step fails
+      fail-fast: false
+      #
+      matrix:
+        repo: ["fsfe/reuse-example", "curl/curl", "SAP/openui5"]
```
carmenbianca, line 27: Good catch on checking the example.

carmenbianca (changes requested):

.github/workflows/3rd_party_lint.yaml
mxmehl: Makes sense. But I wonder where to clone the repos to. Shall we establish the folder "3rd" as an ignored one that can be used for this?

carmenbianca: Maybe using https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory (uses OS default equivalent to `/tmp/[random-hash-or-whatever]/`, but the downside is that typically the directory is deleted when the process ends, which means:

- Can't investigate the directory after the test failed, because the directory is deleted.
- More clones than needed if running successively (but using existing git clones would imply more complex logic; cloning every time is a lot easier).

So maybe `[tests/]3rd/` like you propose.

Alternatively: `~/.cache/reuse/[...]/` (using `appdirs` dependency to find OS-agnostic XDG_CACHE_HOME equivalent).

mxmehl: So I went for the very naive path of just creating a relative directory for these things. A temporary dir would have been more elegant but I figured that this may not work in the Github runner (but I did not try TBH).

mxmehl:
So I'm not sure about the poetry lockfile. I installed poetry 1.1.x as described in the docs but it changed a lot of things in the lockfile that weren't there before, although the newly added dependency didn't introduce them, AFAIK. Not sure how you handled that before.

carmenbianca (commented):
I did some refactoring.

- I reintroduced the matrix so that you can clearly see _which_ repo failed in the CI.
  - There is some duplication here between the workflow file and the Python file to make this work. It's probably fine.
- This requires Poetry 1.2 to make this work because of https://github.com/python-poetry/poetry/issues/2369
  - I will work on making some changes to switch to Poetry 1.2 later.
- Probably some more.

carmenbianca (approved):
LGTM

carmenbianca:
https://invent.kde.org/sysadmin/ci-utilities/raw/master/gitlab-templates/reuse-lint.yml

Apparently KDE removes the po files before doing the check.

carmenbianca:
Should we communicate with the KDE people to see if they can use `.reuse/dep5` to deal with po files?

In any case I would really like to include a KDE project here. I'm happy to implement a hacky workaround to make this test run. Not in reuse itself; just in the test script.

mxmehl:
> Should we communicate with the KDE people to see if they can use `.reuse/dep5` to deal with po files?

Yes, that would be good. Perhaps we can ask @cordlandwehr?

> In any case I would really like to include a KDE project here. I'm happy to implement a hacky workaround to make this test run. Not in reuse itself; just in the test script.

How about we merge this PR without KDE first, and create an issue/PR for the KDE one?

cordlandwehr:
> > Should we communicate with the KDE people to see if they can use `.reuse/dep5` to deal with po files?
> 
> Yes, that would be good. Perhaps we can ask @cordlandwehr?
> 
> > In any case I would really like to include a KDE project here. I'm happy to implement a hacky workaround to make this test run. Not in reuse itself; just in the test script.
> 
> How about we merge this PR without KDE first, and create an issue/PR for the KDE one?

@mxmehl I think our problem in KDE is more a procedure one than a technical one. The story is that many applications were transformed to be REUSE compliment. At that time translations lived outside the repository. About a year ago the translation workflow was changed to move translations into the individual repositories. Yet, most of the are not yet REUSE complient and that will take tike for the reason that translation files are modified by external tools (note: they also write and update copyright information) which are not yet ready to output SPDX complient files.
So, we have a repository that is complient in all parts (and thus shall be covered by the reuse tests) and one folder that is in a kind of staging state and thus need to be excluded for the time being.
Thinking from test systems maybe an option like EXPECTED_FAIL for folders would be helpful here.
The goal in my opinion is to get the PO folders fully REUSE complient though and I do not see any need for dep5, because each file already contains a copyright header.

carmenbianca:
>  The goal in my opinion is to get the PO folders fully REUSE complient though and I do not see any need for dep5, because each file already contains a copyright header.

That's awesome. It depends on the workflow, I think. We use (hosted) weblate, which doesn't add copyright headers.

But if you can implement this without dep5, all the better.

@mxmehl How do you want to proceed? Write a special case for KDE, or create an issue and wait?

mxmehl:
> @mxmehl How do you want to proceed? Write a special case for KDE, or create an issue and wait?

I think it's worth to merge this PR so we can make use of the tests in the upcoming PRs. But yes, please create an issue to add KDE (and perhaps also other repos/orgs), it's worth it :)

carmenbianca (approved):

PR #787, opened 2023-06-25
Make sure the sources are correct when information is aggregated

### LINKED ISSUE #753: Remove potentially superfluous attributes in `_File` class
Since we should have a dataclass containing all the copyright and licensing information when #654 is merged, one might be able to remove the following attributes of the `_File` class:

```python
class _File:
    """Represent an SPDX file. Sufficiently enough for our purposes, in any
    case.
    """
    def __init__(self, name, spdx_id=None, chk_sum=None):
        ...
        self.license_concluded: str = None
        self.copyright: str = None
        ...
```

It is likely that this is a larger refactoring.

carmenbianca:
Since #778, the source information is incorrect. This was flagged before [here](https://github.com/fsfe/reuse-tool/pull/654#discussion_r1161900096).

Also fixes #753

### DIFF
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -92,11 +92,11 @@ class SourceType(Enum):
     """
 
     #: A .license file containing license information.
-    DOT_LICENSE_FILE = ".license file"
+    DOT_LICENSE = ".license file"
     #: A file header containing license information.
     FILE_HEADER = "file header"
     #: A .reuse/dep5 file containing license information.
-    DEP5_FILE = ".reuse/dep5 file"
+    DEP5 = ".reuse/dep5 file"
 
 
 # TODO: In Python 3.10+, add kw_only=True
@@ -107,6 +107,7 @@ class ReuseInfo:
     spdx_expressions: Set[Expression] = field(default_factory=set)
     copyright_lines: Set[str] = field(default_factory=set)
     contributor_lines: Set[str] = field(default_factory=set)
+    path: Optional[str] = None
     source_path: Optional[str] = None
     source_type: Optional[SourceType] = None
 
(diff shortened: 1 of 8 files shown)


src/reuse/__init__.py
```diff
@@ -149,9 +151,16 @@ def union(self, value: "ReuseInfo") -> "ReuseInfo":
                 new_kwargs[key] = attr_val
         return self.__class__(**new_kwargs)  # type: ignore
 
-    def contains_copyright_or_licensing(self) -> bool:
-        """Either *spdx_expressions* or *copyright_lines* is non-empty."""
-        return bool(self.spdx_expressions or self.copyright_lines)
```
carmenbianca, line 154: On second thought, I think we should keep this method (+ its tests)

carmenbianca:
> One small thing I just noticed that should be included here perhaps is an explanation of how this kind of breaks JSON as in the JSON output there really is only one source given.

Au contraire :) Here's a part of the output when running `reuse lint --json` on this repository:

```json
    {
      "path": "po/nl.po",
      "copyrights": [
        {
          "value": "2023 Free Software Foundation Europe e.V. <https://fsfe.org>",
          "source": ".reuse/dep5",
          "source_type": "dep5"
        },
        {
          "value": "SPDX-FileCopyrightText: 2018, 2020 Carmen Bianca Bakker",
          "source": "po/nl.po",
          "source_type": "file-header"
        },
        {
          "value": "SPDX-FileCopyrightText: 2017, 2020 Andr\u00e9 Ockers",
          "source": "po/nl.po",
          "source_type": "file-header"
        }
      ],
      "spdx_expressions": [
        {
          "value": "GPL-3.0-or-later",
          "source": ".reuse/dep5",
          "source_type": "dep5"
        },
        {
          "value": "GPL-3.0-or-later",
          "source": "po/nl.po",
          "source_type": "file-header"
        }
      ]
    },
```

This PR fixes exactly what you just described :D

linozen (approved):

linozen:
> This PR fixes exactly what you just described :D

Amazing! Then this is ready for a merge IMHO.

carmenbianca:
Let's first undo the removal of the method I flagged, and then let's merge :D

PR #791, opened 2023-06-27
msgmerge po files

carmenbianca:
I had suspected that weblate would do this; it doesn't. It only does
this for new translations. This explains why we've received no _new_
translations.


linozen (approved):
LGTM

linozen:
Feel free to merge @carmenbianca

PR #794, opened 2023-07-04
Ignore the LICENCE file

WhyNotHugo:
`reuse lint` complains that my licence file is missing a licence header:

    The following files have no licensing information:
    * LICENCE

Obviously this file doesn't need a licence header, so ignore it entirely.

### DIFF
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -63,6 +63,7 @@ _IGNORE_MESON_PARENT_DIR_PATTERNS = [
 ]
 
 _IGNORE_FILE_PATTERNS = [
+    re.compile(r"^LICENCE"),
     re.compile(r"^LICENSE"),
     re.compile(r"^COPYING"),
     # ".git" as file happens in submodules


nicorikken:
Thanks for raising this. I think we should only consider this in case the LICENCE file is actually a reasonable file for license details and would become part of the REUSE specification. As far as I know only the LICENSE file is actually checked by other tools, so then I don't think this should be part of REUSE.

WhyNotHugo:
> I think we should only consider this in case the LICENCE file is actually a reasonable file for license details

I think that it's a pretty sensible / reasonable place to have a licence. `LICENCE.md` might be common too.

> and would become part of the REUSE specification.

Do you think the specification can be extended to allow this? I'm perfectly aware that `license` is an extremely common usage, despite the typo. Rather than enforce a typo in the spec, I think it makes sense to allow the correct spelling too.

Otherwise we might end up with a situation like the [Referer header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer), where the typo is now part of the spec and if you accidentally use `referrer` (e.g.: no misspelling) things don't work.

carmenbianca:
It's not a typo. It's standard American spelling, which pains me a little, but it is what it is. I can't write `colour` in CSS either, for better or for worse.

This tool used to support both spellings. I've forgotten why the en_GB spelling was dropped. I'd have to research that.

WhyNotHugo:
I've also been experimenting on another project (which includes its licence in a file named `COPYING`, which is rather common).

The file [`COPYING`](https://github.com/pimutils/khal/blob/master/COPYING) contains the MIT licence, but when adding `SPDX-License-Identifier: MIT` to a file, `reuse` complains that the licence is missing:

```
> reuse lint
# MISSING LICENSES

'MIT' found in:
* khal/utils.py
...
```

So it seems that `reuse` checks not only that licences are present, but that a project uses the en_US locale and has a very specific layout for their licences.

Is there a way to _just_ check that SDPX headers are present and that there are no licensing related issues, without trying to enforce a specific project structure and locale?

carmenbianca:
[You need to put the licences in the LICENSES/ directory, even if you only have a single licence](https://reuse.software/faq/#single-license). It does not check the contents of any files to detect whether it matches the licence text of any known licence.

Files like `COPYING` and `LICENSE` are only ignored by the tool to (partially) support old conventions.

> that a project uses the en_US locale

SPDX enforces this too with `SPDX-License-Identifier`. I am not very convinced that en_GB needs to be supported as a special case. I speak five other languages that aren't English, and none of _them_ are supported by any API, programming language, or specification, either. I have some frank thoughts about this, but 'exclusively support en_GB in addition to en_US' is not the solution that speaks to me most.

> Is there a way to just check that SDPX headers are present and that there are no licensing related issues, without trying to enforce a specific project structure and locale?

No.

If you want to do this, you can either write a wrapper around `reuse` that parses the output to check exclusively for what you want (the next release which will contain a `--json` flag might make this much easier), or you can write a program that imports `reuse` as a Python library. Just be warned that we promise no API stability.

But I would recommend full adoption of the REUSE Specification. We're aware that [it breaks with tradition](https://reuse.software/faq/#tradition), but barring some minor details that we're constantly refining, we did put a lot of thought and effort into getting this holistically right.

PR #795, opened 2023-07-04
Fix mypy issue in `docs/conf.py`

WhyNotHugo:
Fixes https://github.com/fsfe/reuse-tool/issues/769


carmenbianca:
Ha, that's one way of fixing it :) I initially wanted to just forcibly ignore this file, but if it was that easy, then I see no reason to not implement this PR.

carmenbianca (approved):

PR #801, opened 2023-07-08
Add Typst file support (.typ)

silverhikari:
Typst is a typesetting program created in rust in the same vein of LaTeX.

the syntax for comments is [here](https://typst.app/docs/reference/syntax/#comments) and uses the same style as the C programming language for both single and multi-line comments, so a new style is not needed.


mxmehl (approved):
LGTM, thank you!

PR #805, opened 2023-07-19
feat: no strict dep5 parsing

### LINKED ISSUE #803: Dep5 Duplicate field "Copyright": Fails to parse many files with REUSE v2.1.0 Release
Hi, in the KDE repos, our CI just picked up the new REUSE release and apparently this introduces a severe regression. Eg. in this job we suddenly have all files not being able to be read: https://invent.kde.org/system/partitionmanager/-/jobs/1064900
Any ideas what might going wrong here?

nicorikken:
Since version 0.1.47 of python-debian the dep5 parsing is strict. This is causing issues for certain users that have multiple Copyright statements in their Dep5 file. Disablng strictness removes the issue for those users.

Closes #803

### DIFF
--- a/src/reuse/project.py
+++ b/src/reuse/project.py
@@ -306,7 +306,7 @@ class Project:
             copyright_path = self.root / ".reuse/dep5"
             try:
                 with copyright_path.open(encoding="utf-8") as fp:
-                    self._copyright_val = Copyright(fp)
+                    self._copyright_val = Copyright(fp, strict=False)
             except OSError:
                 _LOGGER.debug("no .reuse/dep5 file, or could not read it")
             except DebianError:
--- a/src/reuse/report.py
+++ b/src/reuse/report.py
@@ -19,9 +19,9 @@ from pathlib import Path, PurePath
 from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Set, cast
 from uuid import uuid4
 
-from . import __REUSE_version__, __version__
+from . import ReuseInfo, __REUSE_version__, __version__
 from ._util import _LICENSING, StrPath, _checksum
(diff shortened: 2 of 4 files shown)


carmenbianca (changes requested):
This is missing a test. The test can be as simple as 'does not raise an error when trying to parse a (slightly? very?) incorrect DEP5 file'.

nicorikken:
This fails in its current state as multiple copyright fields are simply ignored.

carmenbianca (approved):
I've adjusted the tests a little. No functional changes; just refactored to no longer call to_dict_lint.

carmenbianca:
I am no longer certain if I want to merge this.

I took a look at python-debian 0.1.38 (the lowest compatible version according to pyproject.toml), and strict mode was _always_ enabled.

https://salsa.debian.org/python-debian-team/python-debian/-/blob/0.1.38/lib/debian/copyright.py#L130

I have to assume that python-debian recently included 'no duplicate fields' in their checks, which is fine, but simply disabling this now results in _other_ breakage.

Specifically, the following checks (in 0.1.38, but I'm going to assume that they exist after 0.1.38 also):

```python
                    _complain('Non-header paragraph has neither "Files" nor '
                              '"License" fields', strict)

[...]

            if 'Copyright' not in data:
                _complain('Files paragraph missing Copyright field', strict)
            if 'License' not in data:
                _complain('Files paragraph missing License field', strict)

            if not self.files:
                _complain('Files paragraph has empty Files field', strict)
```

I am fairly certain that we do need/want these checks.

nicorikken:
@carmenbianca Good catch, then merging #808 seems more appropriate.

PR #811, opened 2023-07-27
Link to license list

raboof:
Save users a trip to the search engine

### DIFF
--- a/src/reuse/init.py
+++ b/src/reuse/init.py
@@ -23,7 +23,8 @@ def prompt_licenses(out: IO[str] = sys.stdout) -> List[str]:
     """Prompt the user for a list of licenses."""
     first = _(
         "What license is your project under? "
-        "Provide the SPDX License Identifier."
+        "Provide the SPDX License Identifier. "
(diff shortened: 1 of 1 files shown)


mxmehl (approved):
Good idea, thank you.

src/reuse/init.py
```diff
@@ -23,7 +23,8 @@ def prompt_licenses(out: IO[str] = sys.stdout) -> List[str]:
     """Prompt the user for a list of licenses."""
     first = _(
         "What license is your project under? "
-        "Provide the SPDX License Identifier."
+        "Provide the SPDX License Identifier. "
+        "See https://spdx.org/licenses/ for the list."
```
carmenbianca, line 27:
```suggestion
        "See <https://spdx.org/licenses/> for the list."
```

The added brackets should make it easier to click in some terminals.

raboof: Updated and checked that while the terminal I use (kitty) doesn't need this, it also isn't thrown off by it, so :+1: by me

carmenbianca (approved):
Awesome. Thanks a lot for this contribution @raboof

PR #813, opened 2023-07-28
Add note for translators

carmenbianca:


mxmehl:
Will this be shown in Weblate?

carmenbianca:
yes, that's the idea :)

mxmehl (approved):
Good feature, could also be used for the REUSE website

PR #815, opened 2023-07-31
Add Julia comment style

sorairolake:
Add the comment style for Julia.[^1] This supports both single-line comments and multi-line comments.

Also, single-line multi-line comments cannot be parsed with `CommentStyle.parse_comment()`, so I changed this to run `CommentStyle._parse_comment_multi()` first.

[^1]: https://docs.julialang.org/en/v1/base/punctuation/

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -62,6 +62,8 @@ Contributors
 
 - roberto-red <roberto@redradix.com>
 
+- Shun Sakai
+
 Translators
 -----------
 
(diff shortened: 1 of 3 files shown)


carmenbianca (approved):
Awesome PR. I touched this up a little. I removed the tests because the tests are already generically done for all comment styles. The explicit tests for some other styles are a relic from before I made the tests generic, and they're kept for ease of debugging when something goes horrifically wrong.

Thanks a lot @sorairolake !

PR #820, opened 2023-08-09
Fix reference to spdx/license-list-data

marbre:


mxmehl (approved):
Whoops, good catch, thank you.

marbre:
> Whoops, good catch, thank you.

Thanks for the quick review :)

PR #823, opened 2023-08-20
annotate: add --fallback-dot-license option

anthony-loiseau-act:
`reuse annotate` defaults to fail when it reaches unrecognized files, and to ignore them when `--skip-unrecognised` option is given. When annotating a large set of files at once, creating a .license file for unrecognized files can be handy instead.

This PR creates a new `--fallback-dot-license` option for this case, making it easier to annotate a large repository with various file kinds.

This new option creates a .license file for any non-binary file whose extension is unknown to reuse. Known or binary files are handled as usual. Reminder: files detected as binary are automatically supported with a .license file.

### DIFF
--- a/src/reuse/header.py
+++ b/src/reuse/header.py
@@ -387,11 +387,14 @@ def _get_comment_style(path: StrPath) -> Optional[Type[CommentStyle]]:
     return style
 
 
-def _is_uncommentable(path: Path) -> bool:
+def _is_uncommentable(path: Path, fallback_dot_license: bool) -> bool:
     """Determines if *path* is uncommentable, e.g., the file is a binary or
     registered as an UncommentableCommentStyle.
     """
-    is_uncommentable = _get_comment_style(path) == UncommentableCommentStyle
+    style = _get_comment_style(path)
+    is_uncommentable = style == UncommentableCommentStyle
+    if style is None and fallback_dot_license:
+        is_uncommentable = True
     return is_uncommentable or is_binary(str(path))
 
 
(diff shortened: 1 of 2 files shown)


src/reuse/header.py
```diff
@@ -755,7 +767,7 @@ def run(args: Namespace, project: Project, out: IO[str] = sys.stdout) -> int:
             force_multi=args.multi_line,
         )
         if not args.skip_unrecognised:
```
carmenbianca, line 769: I would remove a lot of logic elsewhere and change this to

```suggestion
        if not args.skip_unrecognised or args.fallback_dot_license:
```

This reduces passing the parcel, and matches where `skip_unrecognised` is used.

src/reuse/header.py
```diff
@@ -613,6 +619,11 @@ def add_arguments(parser: ArgumentParser) -> None:
         choices=list(NAME_STYLE_MAP),
         help=_("comment style to use, optional"),
     )
+    parser.add_argument(
+        "--fallback-dot-license",
+        action="store_true",
+        help=_("Write a .license file for unrecognised files"),
+    )
```
carmenbianca, line 626: Is it possible to use [ArgumentParser.add_mutually_exclusive_group](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) for `--fallback-dot-license`, `--skip-unrecognised`, and `--force-dot-license`?

carmenbianca (changes requested):
Hi @anthony-loiseau-act !

I love this feature proposal a lot. I just have one change I would make in moving some logic, and one improvement that should have been made anyway.

Are you happy to provide a new patch with the feedback? I'm also happy to follow up myself.

anthony-loiseau-act:
I would be pleased to give it a try if delay is not a big matter (since I am not very familiar with python and this code area). I think I can made it within a week max, depending on the time I take to more widely understand involved parts of the code.

carmenbianca:
This is no longer needed with #851 merged. Thank you!

PR #836, opened 2023-10-06
Add modern Fortran comment style

dbroemmel:
The current comment style for Fortran seems to target fixed form source code, where any character in col.1 would start a comment. The character usually is "c" or ''C".  
Free form source code, on the other hand, requires an "!" to start comments. Adding comments with a leading "c" is not supported if I'm not mistaken, sources would no longer compile, `reuse` would thus break the code.

This PR adds an additional comment style for modern free form Fortran and adjusts the selection to commonly used file extensions. While the easiest fix would have been to replace
```python
    SINGLE_LINE = "c"
```
with
```python
    SINGLE_LINE = "!"
```
I think it is more customary to have 'c' with fixed form source code, hence the additional style block.

Since the changes are minimal at best, I did not add a line to "AUTHORS.rst" or the copyright headers at the top of the file. If you prefer to add anything, it should be my employer "Juelich Supercomputing Centre, Forschungszentrum Juelich GmbH", possibly including my name. But I don't think any of this is necessary.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -66,6 +66,8 @@ Contributors
 
 - Shun Sakai
 
+-  Dirk Brömmel
+
 Translators
 -----------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ CLI command and its behaviour. There are no guarantees of stability for the
 
 - More file types are recognised:
   - Julia (`.jl`) (#815)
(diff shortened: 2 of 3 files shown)


carmenbianca (approved):
Hi @dbroemmel ! This PR looks excellent, and thank you for the explanation.

dbroemmel:
Thanks for the merge and amending the authors file!

PR #132, opened 2019-11-30
Support PyPy

carmenbianca:
A tiny side-project. This probably shouldn't be merged until PyPy 3 fixes a bug I opened:

https://bitbucket.org/pypy/pypy/issues/3121/pathlibpath-objects-are-not-accepted-by

PR #839, opened 2023-10-15
.pre-commit-hooks.yaml: remove the language_version

Mikaela:
Reuse specifying `language_version: python3` means that the hook is executed using the system `python3` version and it takes priority over the top level language configuration of the project unlesss the project added explicit override for the reuse hook.

E.g. with:

```yaml
default_language_version:
  python: pypy3
```

all other Python3 `pre-commit` hooks I have would be running with `pypy3` with the exception of `reuse` enforcing `python3`.

Based on https://github.com/fsfe/reuse-tool/pull/132, reuse would support pypy outside of pre-commit.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -10,4 +10,3 @@
   pass_filenames: false
   description:
     "Lint the project directory for compliance with the REUSE Specification"
-  language_version: python3


carmenbianca:
@Mikaela I think we need to make a new release of reuse before this is of any use to you.

Mikaela:
Yes, unless I use the commit hash instead of a tag (and lose the ability to `pre-commit autoupdate`), but I don't think it's so urgent as it will happen eventually with the next release and meanwhile Python 3.12 will work

Mikaela:
I ended up trying this and it works. Pre-commit.ci now reports using pypy3 while with previous commits it reported using python3 in the build section.

- https://results.pre-commit.ci/run/github/397679519/1697781103.spFlDy26QC6RjdJ97lBD5Q

I also didn't lose `pre-commit autoupdate`, it just wants to downgrade to the latest tag, so it will be fine at the time of the next release.

PR #849, opened 2023-10-24
Save the penguins; run fewer tests in CI

carmenbianca:
The biggest change here is that we will no longer be pushing to the dev tag on Docker.


carmenbianca:
Maybe we should also stop building 3 Docker images, and just publish a single image. Probably Debian? Not sure.

People can build their own 'extra' image in any case.

mxmehl:
> Maybe we should also stop building 3 Docker images, and just publish a single image. Probably Debian? Not sure.
> 
> People can build their own 'extra' image in any case.

OK, I see the point of not creating a Docker image for every `dev`, especially in all three versions. Let's drop that.

Regarding `extra`, I'd be fine with dropping it as well, but some people may rely on it. I wonder whether an additional image per release (not commit) makes a huge difference.

Regarding alpine vs. Debian, I strongly suggest to stick with alpine as default and still offer Debian as alternative. If we're speaking about penguins, considering that `fsfe/reuse` has been downloaded more than 1 million times, the lower traffic makes a larger difference than building one more image per release. Debian on the other hand is also important as it's better in terms of license compliance, and some orgs prefer that.

carmenbianca:
@mxmehl full ack, let's do that!

README.md
```diff
@@ -224,14 +224,19 @@ You can also provide additional arguments, like so:
...
-There are a number of tags available:
-
-- `latest` is the most recent stable release.
-- `dev` follows the `main` branch of this repository. Up-to-date, but
-  potentially unstable.
-- `latest-extra` has a few extra packages installed, currently `openssh-client`.
-- `latest-debian` is based on `python:slim`. It is larger, but may be better
-  suited for license compliance.
+Version 2.1.0 and earlier used to have separate Alpine and Debian Docker images.
+Since version 2.2.0, only Debian images are published.
```
mxmehl, line 228: This can probably be removed.

CHANGELOG.md
```diff
@@ -50,6 +50,9 @@ CLI command and its behaviour. There are no guarantees of stability for the
 ### Changed
 
 - Alpine Docker image now uses 3.18 as base. (#846)
+- No longer create and publish `-extra` Docker images. (#849)
```
mxmehl, line 53: Could this even be breaking?

mxmehl: And shall we also mention that the `dev` images will not be supported any more?

README.md
carmenbianca: Good catch.

CHANGELOG.md
carmenbianca: I'm inclined to not treat it as breaking. Nothing breaks; the images will keep working. It's just that no more -extra images will be published.

Added a line for `dev`. Removing `dev` _might_ be a breaking change, but I'm not sure it would matter much whether this is communicated during a minor version or a major version. But honestly, people should expect `dev` to break _anyway_.

mxmehl: OK. Perhaps we should remember ourselves to delete some of these tags altogether so that users know that something broke. Keeping outdated, unversioned tags (like `latest-extra`) around seems like a bad practice.

carmenbianca: Good call, created #850. Thanks a heap for your assistance @mxmehl !

mxmehl (approved):
Thank you!

carmenbianca:
I had another look at #238. Getting rid of `-extra` would (still) basically break CircleCI, which is stupid.

I'm going to make the decision to rope `openssh-client` into the base Alpine image. Being able to `git clone` is a fairly basic functionality that should be available. The difference in size isn't that big:

```
$ podman image ls fsfe/reuse
REPOSITORY            TAG            IMAGE ID      CREATED       SIZE
docker.io/fsfe/reuse  latest-debian  439efbe73d17  3 months ago  340 MB
docker.io/fsfe/reuse  latest-extra   ba02bf7cfdbd  3 months ago  114 MB
docker.io/fsfe/reuse  latest         7cc624f34f51  3 months ago  110 MB
```

(4MB with the latest 2.1 release)

mxmehl:
Sounds fair, thank you for taking an "extra" look ;)

PR #861, opened 2023-11-17
util: fix parsing of copyrights without year

anthony-loiseau-act:
Fixes a crash when reuse tries to parse an older Copyright line which has no year token. This resulted in a None dereference when a string was expected.

### DIFF
--- a/src/reuse/_util.py
+++ b/src/reuse/_util.py
@@ -295,9 +295,9 @@ def _parse_copyright_year(year: str) -> list:
     """Parse copyright years and return list."""
     if not year:
         ret = []
-    if re.match(r"\d{4}$", year):
+    elif re.match(r"\d{4}$", year):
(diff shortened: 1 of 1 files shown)


carmenbianca:
I think I can see what this PR fixes. Do you have an example for how to trigger the bug, though?

Thanks for the patch @anthony-loiseau-act

anthony-loiseau-act:
Sure. I created an  issue ticket to properly sort the root issue I reached and fixed by this proposal. This is #866.

TLDR: Attempting to add a n SPDX  with `reuse annotate --merge-copyrights --license "BSD-3-Clause" --copyright "Foo Bar <foo.bar@example.com>" --year 2023 foo.c` onto a file containing `// Copyright Mr foo` line crashes when parsing year of detected copyright.

PR #862, opened 2023-11-17
comment: add few various file types and names

anthony-loiseau-act:
New files extensions:
- GNU linker scripts (`*.ld`) are commentable as a C file ("// ")
- Assembly source code (`*.s`) is always commentable as a Lisp file ("; ") to my knowledge

New file names:
- `.empty` files may be created in git-like scm having no empty folder support. Given their name and emptiness, those files can be skipped.
- `.shellcheckrc` configuration files are commentable like python files
- the same for in-project `pylintrc` files


carmenbianca:
Thanks @anthony-loiseau-act ! I merged this via another PR to resolve a merge conflict. All-Circuits-Technologies is a GitHub org, which means I couldn't force-push to the branch.

PR #863, opened 2023-11-18
Implement REUSE.toml

### LINKED ISSUE #779: Define an order of precedence + provide a means to override it (REUSE.yaml)
## Help, I got here because I got a PendingDeprecationWarning

> PendingDeprecationWarning: Copyright and licensing information for 'my-project/foo.py' has been found in both 'my-project/foo.py' and in the DEP5 file located at '.reuse/dep5'. The information for these two sources has been aggregated. In the future this behaviour will change, and you will need to explicitly enable aggregation. See <https://github.com/fsfe/reuse-tool/issues/779>. You need do nothing yet. Run with `--suppress-deprecation` to hide this warning.

Don't worry, there's nothing you need to do quite yet, except be vigilant of upcoming changes. 

The reason you're getting this warning is because of the following scenario. You have a file `my-project/foo.py` which contains the following header:

```python
# SPDX-FileCopyrightText: 2023 Jane Doe
#
# SPDX-License-Identifier: GPL-3.0-or-later
```

But you also have a `.reuse/dep5` file which contains the following section:

```
Files: my-project/*.py
Copyright: 2020 Example NGO
License: 0BSD
```

The problem: Under which licence is the file? Who are the copyright holders?

Currently, we err on the side of caution, and just _aggregate_ the results. The answer to both questions is 'both', as far as the tool is concerned.

However, that behaviour is not actually specified in the [REUSE Specification](https://reuse.software/spec/), and there is a consensus among the maintainers of REUSE that the current behaviour isn't _great_. So we want to change it.

## A naïve proposal + some history

We want to define an order of precedence for copyright and licensing information. Here is a concrete proposal:

> Copyright and Licensing Information is considered according to the
> following order of precedence:
>
> 1. Information defined in the `.license` file.
> 2. Information defined in the Commentable File.
> 3. Information defined in `.reuse/dep5`.
>
> There is no merging of information from different sources. Only the
> source with the highest precedence is considered.

In fact, this proposal is so concrete that—for a few hours—it was in REUSE Specification 3.1 and tool version 2.0.0! However, because of quick negative feedback, this update to the specification was promptly reverted, and tool version 2.0.0 was yanked. A little embarrassing on our part, but we're thankful for the constructive feedback.

Copied from the change log:

> While the intention of the breaking change was sound (don't mix information sources; define a single source of truth), there were legitimate use-cases that were broken as a result of this.

The legitimate use-case is the following scenario: You copy a project Foo wholesale into your own project as a static dependency. Foo is not REUSE-compliant, but does contain copyright statements in some code headers. You write a section into `.reuse/dep5` broadly declaring that `static/Foo/*` is under its declared licence, and attribute `The Foo Authors` as the copyright holders. However, because the DEP5 file is now no longer applied to the files that contain copyright statements, REUSE will complain that these files do not have a declared licence.

Within the restrictions of the above proposal, there is no good workable solution to this use-case. You could manually edit the headers (not great, especially when Foo is big, or you regularly need to update it), or you could manually add `.license` files, which may be a huge task.

## An actual but not-so-concrete proposal

We still want to define an order of precedence. But we must provide a way to force aggregation (current behaviour) or hard-coding precedence (e.g., prefer `.reuse/dep5` over the file contents).

There does not yet exist a _concrete_ way of doing this, but you may think of it like this. Given the example `.reuse/dep5` section at the start of this issue, we could instead write this:

```
Files: my-project/*.py
Copyright: 2020 Example NGO
License: 0BSD
Precedence: [file (default)|aggregate|dep5]
```

The problem, however, is that DEP5 does not support this field, and we don't want to make it support this field.

So we want to pivot away from DEP5 and adopt a different configuration method. We've been brainstorming this since 2021 (volunteer projects aren't very fast), and we're internally referring to it as `REUSE.yaml` (although the YAML part is a bit up in the air).

## An actual for-real-this-time concrete proposal

Find below a real and actual concrete proposal:

```toml
# The version of the TOML schema. A simple integer should be fine.
# Mandatory.
version = 1

[[annotations]]
# The path (or paths) that are covered, relative to REUSE.toml's directory. Mandatory.
path = [
     ".bumpversion.cfg",
     "setup.cfg"
]
# A string that defines the precedence of copyright and licensing information.
# The choices are:
#
# - "toml" -> Treat the information in this file as the ultimate authority of
#   the copyright and licensing of the covered files. If multiple nested
#   REUSE.tomls have this precedence for the same file, then the topmost REUSE.toml
#   is authoritative.
# - "closest" -> Use the information closest to the file (including inside the file)
#   if available. If no such information is found, then the information inside this
#   REUSE.toml is applied to that file. TODO: what if there is only partial information
#   inside the file?
# - "aggregate" -> Aggregate the information from this file with the information
#   inside of the covered files.
#
# Not mandatory. Defaults to "closest".
precedence = "toml"
# The copyright notice (or notices) that apply to the above paths. Mandatory.
SPDX-FileCopyrightText = "2017 Free Software Foundation Europe e.V. <https://fsfe.org>"
# The license expression (or expressions) that apply to the above paths.
# Mandatory.
SPDX-License-Identifier = "GPL-3.0-or-later"

# Subsequent tables override previous tables. This does NOT interact with the
# 'precedence' key.
[[annotations]]
# Can contain globs. I would define the globbing behaviour as equal to Python's
# glob.glob (recursive=True, include_hidden=True), but that would require copying
# the glob.glob code and tweaking it to work without filesystem reads. So instead,
# I used the wcmatch.glob 3rd-party library, with GLOBSTAR, MATCHBASE, and DOTGLOB,
# which gives a behaviour very similar to .gitignore and glob.glob (but not precisely
# matching either).
path = "docs/reuse*.rst"
precedence = "toml"
SPDX-FileCopyrightText = [
    "2017 Free Software Foundation Europe e.V. <https://fsfe.org>",
    "2023 Jane Doe",
]
# These SHOULD be joined with AND, but files support multiple separate
# SPDX-License-Identifier tags, so let's support it here as well.
SPDX-License-Identifier = [
    "CC-BY-SA-4.0",
    "GPL-3.0-or-later"
]
```

Some notes on the implementation:

- I chose TOML. I had previously been partial to YAML, but reading over the discussions in the linked issues, and having worked a little more with TOML recently, it's a lot more fool-proof to write, especially as concerns indentation. It's not very good for nesting of data, but we're not doing that, so it's fine. We could bikeshed this choice further, but I propose that we just go with it.
- `path`, `SPDX-FileCopyrightText`, and `SPDX-License-Identifier` can be either a single string or a list of strings. This (partially) matches DEP5 behaviour, making it easy to port. It's also convenient to not _mandate_ lists; we'll probably convert string values into single-value lists in the under-the-hood implementation (edit: that is exactly what I did).
- I chose to use the full `SPDX-[...]` key names. This works better in TOML than in YAML (because the semicolon in YAML messes with this tool's parsing). It's a bit more annoying to type, _but_ it's also very consistent, and means that the user has to memorise less.
- The `version` key doesn't do much of anything precently. I'm not sure if it'll ever become important, but if it does, it'll be good to have.

Some notes about the file itself:

- I intend to support exclusively `REUSE.toml`, and NOT `.REUSE.toml`. People will be peeved by this choice (they don't like random tools littering their clean workspace), but I propose that we stand by this choice. By allowing dotfiles, we would run the risk that the licensing information is hidden on some computers. Licensing information should not be hidden, ergo let's not do dotfiles.
- REUSE.toml files can be nested! You can place them anywhere in the project, not just in the root directory. They use the same `closest`/`aggregate`/`toml` precedence system. `closest` resolves to the file itself OR to the nearest REUSE.toml (can be self). `aggregate` just aggregates the REUSE.toml's information always, and then behaves like `closest`. `toml` is an aggressive "ignore everything underneath me; I am the ultimate authority here" precedence setting. The topmost REUSE.toml with `toml` is authoritative.

## Related issues

Here are some issues of relevance (in order of relevance, feel free to reference more):

- https://github.com/fsfe/reuse-docs/issues/81
- https://github.com/fsfe/reuse-docs/issues/70
- https://github.com/fsfe/reuse-tool/issues/775
- https://github.com/fsfe/reuse-tool/issues/572
- https://github.com/fsfe/reuse-tool/issues/661

This issue will exist as a sort of meta issue to refer back to and track work in other issues.

carmenbianca:
Fixes #779 

See that issue for a more thorough explanation.


carmenbianca:
Some notes to self:

- Test GlobalLicensingConflict
- Test that all the new errors (especially validation errors) have nice user-friendly error messages
- More tests
- More documentation

src/reuse/_main.py
```diff
@@ -296,7 +295,7 @@ def main(args: Optional[List[str]] = None, out: IO[str] = sys.stdout) -> int:
                 dep5=root / ".reuse/dep5"
             )
         )
-    except DebianError as error:
+    except GlobalLicensingParseError as error:
         main_parser.error(
             _(
                 "'{dep5}' could not be parsed. We received the following error"
```
carmenbianca, line 301: this doesn't seem right

src/reuse/global_licensing.py
```diff
@@ -0,0 +1,562 @@
...
+        # Sort from topmost to deepest directory.
+        found.sort(key=lambda toml: toml.directory.parts)
+        return found
+
+    def _find_relevant_tomls_and_items(
+        self, path: StrPath
+    ) -> List[Tuple[ReuseTOML, AnnotationsItem]]:
+        # *path* is relative to the Project root, which is the *source* of
+        # *NestedReuseTOML, which itself is a relative (to CWD) or absolute
+        # *path.
```
carmenbianca, line 551: remove extraneous *s

carmenbianca:
Notes to self:

- Add FORCEUNIX to wcmatch
- Add a test whether a REUSE.toml file can match a file that is above itself in the hierarchy (single file name)

perillo:
What about renaming `REUSE.toml` to `license.toml`?

Thanks.

carmenbianca:
Another note to self: see how this meshes with SPDX-FileContributor

src/reuse/global_licensing.py
```diff
@@ -0,0 +1,536 @@
...
+
+    paths: Set[str] = attrs.field(
+        converter=_str_to_set, validator=_validate_set_of_str
+    )
+    copyright_lines: Set[str] = attrs.field(
+        converter=_str_to_set, validator=_validate_set_of_str
+    )
+    spdx_expressions: Set[Expression] = attrs.field(
+        converter=_str_to_set_of_expr, validator=_validate_set_of_expr
+    )
```
carmenbianca, line 329: As-is, these validators are also valid for zero-length sets, meaning that the following REUSE.toml would be 'valid':

```toml
version = 1
[[annotations]]
path = []
SPDX-FileCopyrightText = []
SPDX-License-Identifier = []
precedence = "toml"
```

This is _probably_ fine if we don't forbid zero-length sets in the specification. It's just … meaningless, especially if `path` is empty.

I'm going to see about changing this a little to avoid some awkwardness here, but no solution is going to be perfect.

carmenbianca: - Forbade path from being zero-length.
- At least one of SPDX-FileCopyrightText and SPDX-License-Identifier must now be non-empty (or even defined).

carmenbianca:
Marking this ready for review.

REUSE.toml
```diff
@@ -0,0 +1,56 @@
...
+version = 1
+
+[[annotations]]
+path = [
+     ".bumpversion.cfg",
+     "setup.cfg"
+]
+precedence = "override"
+SPDX-FileCopyrightText = "2017 Free Software Foundation Europe e.V. <https://fsfe.org>"
+SPDX-License-Identifier = "GPL-3.0-or-later"
```
Kladki, line 14: I am not that big of a fan of how verbose these keys are, and how they don't follow the style of the other keys. Maybe something like this would make more sense?
```toml
copyright_text = "2017 Free Software Foundation Europe e.V. <https://fsfe.org>"
license_identifier = "GPL-3.0-or-later"
```

Kladki: Or should I move this concern/comment to https://github.com/fsfe/reuse-website?

PR #867, opened 2023-11-24
Ignore .sl directory the same way as .hg.

Flameeyes:
The .sl directory is used by Sapling (https://sapling-scm.com/) so
it should be handled like the other SCM directories.


carmenbianca:
Thanks for the PR @Flameeyes 

Can you add a comment explaining what the `.sl` directory is and add an entry in the change log?

Flameeyes:
Done!

Issue #881, opened 2023-12-15
Add option to pay attention to files named `COPYING` or `LICENSE`


Jayman2000:
The REUSE Tool ignores files whose names start with `LICENSE` or `COPYING`. I tried creating a repo like [this one](https://github.com/Jayman2000/reuse-ignores-copying) that contains a `COPYING.md` file that just says that the repo is REUSE compliant. When I run `reuse lint`, it tells me that `CC0-1.0` is unused despite the fact that `COPYING.md` contains `SPDX-License-Identifier: CC0-1.0`:

```console
$ cd reuse-ignores-copying
$ cat COPYING.md
<!--
SPDX-License-Identifier: CC0-1.0
SPDX-FileCopyrightText: 2023 Jason Yundt <jason@jasonyundt.email>
-->

# Copying Information

This repo is [REUSE](https://reuse.software) compliant.
$ reuse lint
# UNUSED LICENSES

The following licenses are not used:
* CC0-1.0



# SUMMARY

* Bad licenses: 0
* Deprecated licenses: 0
* Licenses without file extension: 0
* Missing licenses: 0
* Unused licenses: CC0-1.0
* Used licenses: 0
* Read errors: 0
* files with copyright information: 0 / 0
* files with license information: 0 / 0

Unfortunately, your project is not compliant with version 3.0 of the REUSE Specification :-(
$ 
```

---

As far as I know, this is intentional. According to [this comment](https://github.com/fsfe/reuse-tool/issues/286#issuecomment-746258563) from @mxmehl,

> Ah, I didn't see the LICENSE.md file. Indeed, it's not checked by the tool because it's considered to contain a license text (which per definition is excluded from the obligation to contain copyright and licensing info).

Max is right. [The spec says](https://github.com/fsfe/reuse-docs/blob/0913b0a83b36c161966be1c5e70c81bdadfb8a69/spec.md#definitions)

> - Covered File --- any file in a Project, except for
>     - The License Files.
>     - […]

[and](https://github.com/fsfe/reuse-docs/blob/0913b0a83b36c161966be1c5e70c81bdadfb8a69/spec.md#copyright-and-licensing-information)

> Each Covered File MUST have Copyright and Licensing Information associated with
> it.

[...]

PR #886, opened 2024-01-04
Ignore COPYING and LICENSE as exact matches or with extensions

carmenbianca:
Related to #881 and https://github.com/fsfe/reuse-docs/issues/139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@ CLI command and its behaviour. There are no guarantees of stability for the
 ### Changed
 
 - `.s` files now use the Python comment style as per GNU Assembler (gas). (#928)
+- Previously, any file that begins with `COPYING` or `LICENSE` was ignored. This
+  has been changed. Now, files like `COPYING_README` are no longer ignored, but
+  `COPYING` and `COPYING.txt` are still ignored (in other words: exact matches,
+  or `COPYING` + a file extension). Idem ditto for `LICENSE`. (#886)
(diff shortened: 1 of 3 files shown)


mxmehl:
Uh, this will break things. I see many LICENSE.md or LICENSE.txt files around.

I don't quite understand why this change is needed. Do you want to avoid that a file such as `LICENSE-finder.py` is ignored? In this case, we could find a more elaborate regex.

carmenbianca:
> Uh, this will break things. I see many LICENSE.md or LICENSE.txt files around.

Good shout. I changed this so that `COPYING` and `COPYING[.any_ext_here]` are ignored (ditto for `LICENSE`), but `COPYING_foobar` is not ignored.

CHANGELOG.md
```diff
@@ -67,6 +67,10 @@ CLI command and its behaviour. There are no guarantees of stability for the
 
 ### Changed
 
+- Previously, any file that begins with `COPYING` or `LICENSE` was ignored. This
+  has been changed. Now, files like `COPYING_README` are no longer ignored, but
+  `COPYING` and `COPYING.txt` are still ignored (in other words: exact matches,
+  or `COPYING` + a file extension). Idem ditto for `LICENSE`. (#886)
```
carmenbianca, line 73: This is now in the 3.0.0 change log section…

mxmehl (approved):
Thanks for fixing this issue :)

PR #858, opened 2023-10-31
Add support for the Pijul VCS

korrat:
[Pijul](https://pijul.org) is a new VCS based on a theory of commutative patches. This commit adds support for Pijul to reuse-tool by implementing `VCSStrategyPijul`.

PR #899, opened 2024-01-19
Add support for Sapling as a VCS as well.

Flameeyes:
Add support for Sapling as a VCS as well.

I introduced support to ignore the `.sl` directory already, but that did
not include support for the VCS interface.

As a VCS, Sapling uses effectively the same interface as Mercurial,
although notably it does not support Mercurial's `.hgignore` file, nor
does it invent a new one: it uses `.gitignore`.

I opted for copying the Mercurial implementation though, it looks
more adequate.


src/reuse/vcs.py
```diff
@@ -249,6 +249,76 @@ def find_root(cls, cwd: Optional[StrPath] = None) -> Optional[Path]:
...
+        return path in self._all_ignored_files
+
+    def is_submodule(self, path: StrPath) -> bool:
+        # TODO: Implement me.
+        return False
+
+    @classmethod
+    def in_repo(cls, directory: StrPath) -> bool:
+        if directory is None:
+            directory = Path.cwd()
```
carmenbianca, line 294: This segment of code isn't needed. I'll remove it from the other methods too.

carmenbianca (changes requested):
Hi @Flameeyes! Thanks for the pull request. This looks pretty good, but is missing some stuff. The #858 PR is a good example of things that need to be added:

- A change log entry
- A mention in README
- A small collection of tests
- A workflow where Sapling is installed and tested

src/reuse/vcs.py
carmenbianca: Done in #969

Issue #578, opened 2022-08-25
When running reuse via pre-commit, ignore untracked files (or: create `lint-file`)


vHanda:
Hello.

I sometimes have some random files which I haven't committed and don't necessarily plan to, but I don't want to add them to the '.gitignore'. Reuse complains about those files.

It would be awesome if I could run reuse with a `--ignore-untracked` option, which ignores all files which aren't in git index. 

---

comment by @carmenbianca :

> Other linters, pylint etc, when run via pre-commit, only run against staged/tracked files. This is a behaviour of pre-commit, which passes (I think) a list of files to the linters. reuse differs here, because there's presently no way to run the linter against a subset of files.

Let's create `lint-file` to lint a subset of files, and then allow the pre-commit to be configured to use that instead of `lint`.

carmenbianca:
Tempted to mark this wontfix as feature creep. `pre-commit` can handle this for you. When you commit some files using pre-commit, it quickly stashes the untracked files, runs the pre-commit stuff, then un-stashes the untracked files.

carmenbianca:
Tempted to mark this wontfix as feature creep. `pre-commit` can handle this for you. When you commit some files using pre-commit, it quickly stashes the untracked files, runs the pre-commit stuff, then un-stashes the untracked files.

ryandesign:
Fine, then please update the documentation to provide the complete set of steps needed to accomplish this. Assume I didn't know anything about pre-commit before discovering the reuse tool.

ryandesign:
To clarify: I have set up pre-commit to the best of my ability following the directions currently in the reuse documentation. Pre-commit runs the reuse tool when I commit anything. I prevents any commit from going through if there are any files in the directory that don't have proper license information, even if those files are untracked and are not part of the commit.

carmenbianca:
Ah, you may be right. pre-commit stashes and unstashes unstaged tracked files.

We'll need to look at this issue some time then. I'm inclined towards a solution that doesn't introduce git-specific interactions into the tool. I think the solution might be to only lint the files tracked by pre-commit. There's an issue open for only linting specified files, but I'm on my phone and can't easily find it.

Wuestengecko:
@carmenbianca

> I'm inclined towards a solution that doesn't introduce git-specific interactions into the tool.

It has those already:

[...]

PR #903, opened 2024-02-05
Add `lint --only <files...>` option

Wuestengecko:
Here's my second shot at #578. I went with the `--only` option that @stephanlachnit proposed, as I found it a bit easier to implement than a new subcommand. It should also be fairly simple to extend it with an `--exclude` option later.

---

This option restricts the generated report to only the files listed on the command line. It is useful for example in the pre-commit hook, in order to not fail the entire commit if there are some issues in files that aren't supposed to be committed yet anyways.

This option is implemented by checking all files that we know about as usual, but only reporting on (and failing for) issues in files listed.

We have to check all files, otherwise we wouldn't be able to catch certain types of errors reliably or at all. Specifically, the "unused licenses" check would often raise false-positive errors: If we only look at the *to-be-committed* files, and those don't use a particular license, we would fail the check even if another file uses it (which was already committed before, but not touched during the new commit).

This does however mean that untracked files could cause false-negatives, but this was already the case before, and will (still) be caught by CI.

One alternative to this approach is to not check for unused licenses at all if doing such a partial check, but this would mean that downstream CI scripts would have to be adjusted to explicitly run a full `reuse lint` instead of running the pre-commit hook over all repo files. However, this would only lead to a lot of complications without any meaningful benefit.

---

Annoyingly pylint now complains about "too many locals", because I introduced a new argument to `ProjectReport.generate`, which ruins the otherwise perfect score. To avoid this, I have renamed the `missing_license` and `bad_license` loop variables to `license_name`, because they're both just strings anyways (from a type checker's perspective), and IMO license_name is still explicit enough. It's a separate commit, so if you don't want that, we can just force-push it away and do something else instead :)

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -5,8 +5,8 @@
 - id: reuse
   name: reuse
   entry: reuse
-  args: ["lint"]
+  args: ["lint", "--only"]
   language: python
-  pass_filenames: false
+  require_serial: true
   description:
     "Lint the project directory for compliance with the REUSE Specification"
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2017 Free Software Foundation Europe e.V. <https://fsfe.org>
 # SPDX-FileCopyrightText: 2022 Florian Snow <florian@familysnow.net>
 # SPDX-FileCopyrightText: 2023 DB Systel GmbH
+# SPDX-FileCopyrightText: 2024 Martin Lehmann <lemmi59612@gmail.com>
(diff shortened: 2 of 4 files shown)


Wuestengecko:
Hey! It's been quite a while since I opened this PR. Has anyone had a chance to look at it already? It seems that community interest is definitely there, and I don't want to see it just die completely.

Wuestengecko:
Ah, that's good to know, thanks for the info. I'll try to change this into a subcommand one of these days, but I can't guarantee that's going to happen before you get to it :)

carmenbianca:
Closed by #1055, thank you!

Wuestengecko:
Great, thank you too!

PR #909, opened 2024-02-14
--style should work with unrecognised files

### LINKED ISSUE #902: Forcing a comment style seems broken, support for Cargo.lock
Forcing a style with ```--style [style]``` seems to be broken in version 3.0.1. The error message informing me that a file type does not have a recognised file extension is the same whether this option is specified or not.

On another note, Cargo.lock files are currently not recognised (which is how I discovered the bug above). There was issue #390 that dealt with the topic, recommending either dot licenses or python-style comments. I would be open to implementing that, but there probably should be a discussion which of these styles should be used. Personally, I would prefer the in-file python style.

### LINKED ISSUE #917: `reuse annotate` doesn’t use `--style` if file extension isn’t recognised
## Test case
With `reuse init` already run and where "CC0-1.0" is one of the used licenses:
```sh
touch test.txt
reuse annotate --copyright "me!" --license CC0-1.0 --style python ./test.txt
```

## Expected outcome
`test.txt` would have a Python style header indicating "me!" as the author and CC0-1.0 as the license.

## Actual outcome
```
reuse annotate: error: The following files do not have a recognised file extension. Please use --style, --force-dot-license, --fallback-dot-license, or --skip-unrecognised:

test.txt
```
Nothing is written to `test.txt`

## Enviroment
```
> reuse --version
reuse 3.0.1
> distro
Name: Arch Linux
Version: rolling (n/a)
Codename: n/a
```

carmenbianca:
Fixes #902
Fixes #917

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,9 @@ CLI command and its behaviour. There are no guarantees of stability for the
 
 ### Fixed
 
+- `annotate`'s '`--style` now works again when used for a file with an
+  unrecognised extension. (#909)
+
 ### Security
 
 ## 3.0.1 - 2024-01-19
--- a/src/reuse/_annotate.py
+++ b/src/reuse/_annotate.py
@@ -342,11 +342,13 @@ def verify_write_access(
 
 
 def verify_paths_comment_style(args: Namespace, paths: Iterable[Path]) -> None:
-    """Exit if --fallback-dot-license or --skip-unrecognised is not enabled and
-    one of the paths has an unrecognised style.
+    """Exit if --style, --force-dot-license, --fallback-dot-license,
+    or --skip-unrecognised is not enabled and one of the paths has an
+    unrecognised style.
     """
     if (
-        not args.fallback_dot_license
+        not args.style
+        and not args.fallback_dot_license
         and not args.skip_unrecognised
         and not args.force_dot_license
     ):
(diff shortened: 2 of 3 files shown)


lpechacek (approved):
I hit the bug as well and arrived at roughly the same solution. FWIW, I think that the fix is correct.

For the record, my fix
```
From 47a9de42353f680c9d62550c02f24fe17c5982e5 Mon Sep 17 00:00:00 2001
From: Libor Pechacek <lpechacek@gmx.com>
Date: Mon, 1 Apr 2024 12:13:29 +0200
Subject: [PATCH 1/2] Make the --style option work again

---
 src/reuse/_annotate.py | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/src/reuse/_annotate.py b/src/reuse/_annotate.py
index b867d62a2faf..07d892ad8bbf 100644
--- a/src/reuse/_annotate.py
+++ b/src/reuse/_annotate.py
@@ -492,7 +492,8 @@ def run(args: Namespace, project: Project, out: IO[str] = sys.stdout) -> int:
 
     paths = all_paths(args, project)
 
-    verify_paths_comment_style(args, paths)
+    if args.style is None:
+        verify_paths_comment_style(args, paths)
 
     if not args.force_dot_license:
         verify_write_access(paths, args.parser)
-- 
2.44.0
```

PR #928, opened 2024-03-04
Assume  's' and 'S' sufixes for GNU Assembler, and 'asm' for NASM

### LINKED ISSUE #927: Review assembly comment default syntax
The suffix `.s` is the default extension for assembly code used by the GNU tool-chain. GCC outputs `foo.s` out of `foo.c` if the command-line option `-S` is selected. The GNU Assembler (GAS), which reads the `.s` file as input, expects Python-style comments  (in x86 platform). Also, GAS manual recommends `.s` (lowercase) for compiler-generated code, and `.S` (uppercase) for human-written code.

On the other hand, the suffix `.asm` is commonly used for assembly code meant for the NASM assembler, which uses Lisp-style comments.

Currently, `reuse` treats `.s` files as if they were in NASM syntax, and by add lisp-style comments,  causes  projects that rely on the GNU tools fail to build.

Suggestion: it might be more intuitive and practical for most programmers if 

- `.s` and `S` were reserved for GCC/GAS asembly (usually AT&T syntax), Python-style comment
- `.asm` were reserved for NASM assembly code (Intel syntax), Lisp-style comment

monacofj:
Fixes #927.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -73,7 +73,7 @@ jobs:
       - name: Test formatting with black
         run: |
           poetry run isort --check src/ tests/
-          poetry run black .
+          poetry run black --check .
 
   mypy:
     runs-on: ubuntu-22.04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,25 @@ CLI command and its behaviour. There are no guarantees of stability for the
 ### Security
 -->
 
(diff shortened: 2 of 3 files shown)


carmenbianca:
Thank you @monacofj for your contribution! I have patched it up a little bit, and now it's ready to merge. I also spotted and corrected a little bug thanks to your contribution, which is great news.

PR #938, opened 2024-03-26
Add support for CUDA-C++

Gusser93:
This PR adds ´.cu´ and ´.cuh´ to the list of supported files. As an extension to C/C++, CUDA-C++ uses the same comment style as C/C++.


carmenbianca (approved):
LGTM, thanks a lot for the contribution @Gusser93 !

Issue #384, opened 2021-06-16
`addheader` puts header before the BOM in UTF-8 + BOM files


rpavlik:
Microsoft tools tend to insert byte-order marks fairly liberally, and it seems that however we're loading the file with addheader, it's not "dealing with" that BOM correctly: the header gets inserted before the bom resulting in something that looks like this:

![image](https://user-images.githubusercontent.com/61129/122255903-34212080-ce94-11eb-900b-43e8559e997f.png)

This is the file I was adding a header to: https://github.com/walbourn/directx-vs-templates/blob/master/VSIX/src/Wizards/MyPlatformToolset.cs

mxmehl:
Oh, I hate encoding so much. So the same problem would also arise with UTF-16 and UTF-32 where BOMs are even required, right?

Would it suffice if `addheader` would detect BOMs and insert its part after the byte order mark?

rpavlik:
I haven't dared try utf-16 text files since I try to avoid them but probably it would respond badly, possibly. Or it may just be that utf-8 boms (which are optional) are just handled badly by python, which would put it in good company (Most non-MS-focused tools stumble on them).

Yes, that's probably enough.

mxmehl:
OK, sounds like a plan:
1. Detect the most common BOMs
2. If any is present, add stuff after BOM. If not, proceed as usual.
3. Test things thoroughly.

dkfellows:
This would be a trivial problem to fix except you really want the change to be BOM-preserving. That makes it a bit messier, as that's information that Python's `utf_8_sig` codec doesn't expose. Our fix is in https://github.com/UoMResearchIT/reuse-tool/pull/6 (our tests aren't public)

PR #940, opened 2024-03-29
Add support for certain files from .NET ecosystem

ForNeVeR:
- Visual Studio solution file, `.sln`
- .NET project files: `.csproj`, `.fsproj`, `.vbproj`
- F# script files, `.fsx`
- MSBuild files, `.props` (such as `Directory.Build.props`)

While working with the tool, I've also noted that it can sometimes break files if they include UTF-8 BOM (which is, while not mandatory, pretty common in .NET ecosystem). This is a separate issue #384, though.


carmenbianca:
Thanks @ForNeVeR for the patch! It looks great, merging this.

PR #941, opened 2024-03-30
comment: C, Cpp and Css styles reorganization (big)

Thannoy:
C and C++ features different comment styles:
- only multi-line `/* */` comment is valid in traditional C89
- C++ also supports `// ` single line comments from its beginning
- C started to support C++-style `// ` comments in C99

For this reason, multi-line `/* */` is often named C-style comment and single-line `// ` is often named C++-style comment.

In current implementation, there were only a "CCommentStyle" with `/* */` and `// ` support, used for both C and C++ files (also used by quite many other file types). 

Therefore:
- Rename CCommentStyle as CppCommentStyle
- Create a new CCommentStyle with only `/* */` support
- Remove redundant CssCommentStyle (same as new CCommentStyle)
- Associate *.c and *.h with CCommentStyle

Caution:
- this (slightly) changes *.c and *.h comments (by removing their multi-line support)
   - despite *.h files being often used also for C++, which is a minor issue in my opinion since C comments is a subset of C++ comment
- this changes the list of known style shorthands:
   - "c" style shorthand meaning changes
   - "css" style shorthand is no more known (want we keep a redundant CssCommentStyle to avoid this?)

This PR is a little bit big but in my opinion results in a clearer handling of C/C++ (and Css).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,12 +70,13 @@ CLI command and its behaviour. There are no guarantees of stability for the
 
 ### Deprecated
 
+- `csingle` and `css` style shorthands (see changes). (#941)
+
 ### Removed
 
 - The PendingDeprecationWarning for the aggregation of information between DEP5
   and the contents of a file has been removed. This behaviour is now explicitly
   specified in REUSE Specification v3.2. (#1017, related to #779)
-- `csingle` and `css` style shorthands (see changes). (#941)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,8 +55,18 @@ CLI command and its behaviour. There are no guarantees of stability for the
 - Support alternate spelling `--skip-unrecognized`. (#974)
 - In `annotate`, rename `--copyright-style` to `--copyright-prefix`. The former
   parameter is still supported. (#973)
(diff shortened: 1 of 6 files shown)


Thannoy:
Another caution about this PR: since it touches quite many places, conflicts with other PRs are likely.

Thannoy:
- Resync'ed with main 1a131cc
- added missing cppsingle shorthand update

Thannoy:
For information purpose, resyncing this PR with master updates is not that trivial. Here is my workflow:
- git fetch
- git merge origin/main
- re-do comment.py d073a5 key modifications:
  - git checkout origin/main -- src/reuse/comment.py
  - rename all occurrences of CCommentStyle to CppCommentStyle
  - rename all occurrences of CssCommentStyle to CCommentStyle
  - fix impacted *CommentStyle themselves and their shorthands (C/Cpp/CppSingle/Css)
  - change .c and .h comment style to CCommentStyle
- tests

carmenbianca:
Hi @Thannoy and thanks! I rebased on top of master again, and did some more adjusting. To maintain support for the old shorthands, I manually added them to the dictionary and added tests for this.

This PR is _technically_ a breaking change still, but I can live with that.

Thanks for the work and the history context.

PR #944, opened 2024-04-07
Add optional parameters for reuse-init

### LINKED ISSUE #943: Add arguments for `reuse init`
When running `reuse init`, I'd like to be able to pass options with only missing values prompted.  For example:

```sh
$ reuse init --license MPL-2.0 --license CC-BY-SA-4.0
Initializing project for REUSE.

Chosen licenses:
MPL-2.0
CC-BY-SA-4.0

What is the name of the project?
```

The reason for this is so that the init command can be aliased or scripted to something short and the developer doesn't have to repeat values. For example, I'd like to create a shell script for the init command like this:

```bash
project-init() {
  name=$(basename $PWD)
  reuse init --license MPL-2.0 --license CC-BY-SA-4.0 \
    --maintainer "$(git config user.name)" \
    --email "$(git config user.email)" \
    --project-name "$name" \
    --project-address "https://example.com/projects/$name"
}
```

It's not able to create such script with only interactive prompt as of now.

It would be also nice to have environment variable used to set default values that can be overwritten by the command line arguments.

Huy-Ngo:
Resolve #943

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -80,6 +80,8 @@ Contributors
 
 - Apteryks
 
+- Ngô Ngọc Đức Huy <huyngo@disroot.org>
+
 Translators
 -----------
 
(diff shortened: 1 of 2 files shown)


Huy-Ngo:
Closing as #943 is no longer relevant

Issue #925, opened 2024-02-28
Add a simplified machine parsable output for `reuse lint` 


nogweii:
A middle ground between the full JSON output and the human-focused plain output. My idea here is integrating `reuse lint` with tools like [reviewdog](https://github.com/reviewdog/reviewdog) where it excels with output along the lines of:

```
scripts/foo1.sh: missing license 'MPL-2.0'
scripts/foo2.sh: no licensing information
scripts/foo3.sh: no copyright information
```

One can generate such an output using the JSON version and processing it, but I think it would be beneficial to include it upstream.

KlfJoat:
+1 to this. I'd also suggest allowing us to pick from a few different output formats. And those output formats should be existing output formats.

One of the output formats that's simplified and machine-parseable could be [TAP format](https://testanything.org/). Since TAP parsers abound, the direct output of `reuse lint` could be used in CI/CD pipelines.

Another format, which wouldn't be simplified and human-readable, but is widely used could be [JUnit format](https://www.ibm.com/docs/en/developer-for-zos/16.0?topic=formats-junit-xml-format), which also has a lot of parsers and could be consumed by CI/CD pipelines, including [built-in reporting on sites like GitLab](https://docs.gitlab.com/ee/ci/testing/unit_test_reports.html#how-to-set-it-up).

carmenbianca:
Thanks for the issue. This seems quite doable to me. Off the top of my head, an implementation would need:

- A new function akin to `format_plain` and `format_json` in `lint.py`.
- A new argument that can be passed to `reuse lint`, à la `reuse lint --tap` or some such.
- `--json` and `--tap` must be mutually exclusive. `argparse` already supports something like that.
- Some tests.

The TAP format looks decent to me as far as machine-readable specs go. A vague proposal:

```
1..n
ok - ./CHANGELOG.md
ok - ./README.md
not ok - ./pyproject.toml
  ---
  failed:
    - no-copyright
    - no-license
  ...
# global tests
ok - no missing licenses
ok - no unused licenses
ok - no bad licenses
not ok - no licenses without extension
  ---
  failed:
    - ./LICENSES/MIT
  ...
ok - no read errors
```

The YAML is a bit verbose, but there's no other way to document why the test failed.

Figuring out the value of `n` in advance may be a little tricky. I suppose it's `number of files (computed in advance) + number of global tests (static)`.

carmenbianca:
[...]

PR #956, opened 2024-04-10
feat: lint output per line

nicorikken:
Adds an '-line' or '-l' option to the 'lint' command. Prints a line for
each error, starting with the file to which the error belongs.

This output can be a starting point for some parsers, in particular ones
that implement something similar to Vim errorformat parsing.

Related to https://github.com/fsfe/reuse-tool/issues/925, needed for #578

Additional work needed:

- [x] Needs tests
- [ ] Error messages might have to be improved
- [x] Not all errors are found, as some issues aren't in the
  FileReports. This requires additional investigation.

Signed-off-by: Nico Rikken <nico@nicorikken.eu>

### DIFF
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -305,6 +305,13 @@ def format_lines(report: ProjectReport) -> str:
                 )
             )
 
+        # Unused licenses
+        for lic in sorted(report.unused_licenses):
+            lic_path = license_path(lic)
+            output.write(
+                _("{lic_path}: unused license\n").format(lic_path=lic_path)
+            )
+
         # Missing licenses
         for lic, files in sorted(report.missing_licenses.items()):
             for path in sorted(files):
@@ -313,13 +320,6 @@ def format_lines(report: ProjectReport) -> str:
                         path=path, lic=lic
                     )
                 )
-
-        # Unused licenses
-        for lic in sorted(report.unused_licenses):
-            lic_path = license_path(lic)
-            output.write(
-                _("{lic_path}: unused license\n").format(lic_path=lic_path)
-            )
 
         # Read errors
         for path in sorted(report.read_errors):
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -285,7 +285,7 @@ def format_lines(report: ProjectReport) -> str:
         # Bad licenses
         for lic, files in sorted(report.bad_licenses.items()):
             for file in sorted(files):
-                output.write(_(f"{file}: bad license: {lic}\n"))
+                output.write(_(f"{file}: bad license {lic}\n"))
 
         # Deprecated licenses
         for lic in sorted(report.deprecated_licenses):
@@ -300,7 +300,7 @@ def format_lines(report: ProjectReport) -> str:
         # Missing licenses
         for lic, files in sorted(report.missing_licenses.items()):
             for file in sorted(files):
-                output.write(_(f"{file}: missing license: {lic}\n"))
+                output.write(_(f"{file}: missing license {lic}\n"))
 
         # Unused licenses
         for lic in sorted(report.unused_licenses):
--- a/src/reuse/lint.py
+++ b/src/reuse/lint.py
@@ -19,7 +19,7 @@ from typing import IO, Any
 
 from . import __REUSE_version__
 from .project import Project
-from .report import FileReport, ProjectReport
+from .report import ProjectReport
 
 
 def add_arguments(parser: ArgumentParser) -> None:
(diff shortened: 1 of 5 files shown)


nicorikken:
Current version works based on ProjectReport to ensure also license errors are included. A structural improvement might be to create structural reports for license errors, even as FileReports.

nicorikken:
Two remaining suggestions for the future:

- Include source code line numbers of the finding. Requires keeping track of this information when detecting
- How best to deal with edge-case filenames that include colons and spaces?

Current output:

```
/tmp/pytest-of-nico/pytest-44/fake_repository14/invalid-license.py: bad license: invalid
LICENSES/invalid-license-text: bad license: invalid-license-text
LICENSES/Nokia-Qt-exception-1.1.txt: deprecated license
LICENSES/MIT: license without file extension
/tmp/pytest-of-nico/pytest-44/fake_repository14/invalid-license.py: missing license: invalid
LICENSES/MIT: unused license
LICENSES/Nokia-Qt-exception-1.1.txt: unused license
LICENSES/invalid-license-text: unused license
/tmp/pytest-of-nico/pytest-44/fake_repository14/restricted.py: read error
/tmp/pytest-of-nico/pytest-44/fake_repository14/file with spaces.py: without license
/tmp/pytest-of-nico/pytest-44/fake_repository14/file:with:colons.py: without license
/tmp/pytest-of-nico/pytest-44/fake_repository14/no-license.py: without license
/tmp/pytest-of-nico/pytest-44/fake_repository14/file with spaces.py: without copyright
/tmp/pytest-of-nico/pytest-44/fake_repository14/file:with:colons.py: without copyright
/tmp/pytest-of-nico/pytest-44/fake_repository14/invalid-license.py: without copyright
```

nicorikken:
I'm now trying to validate with the [errorformat library](https://github.com/reviewdog/errorformat): `errorformat "%f: %m"`
It seems it cannot succesfully parse the output of files with spaces and colons:
```
/tmp/pytest-of-nico/pytest-44/fake_repository14/invalid-license.py|| bad license invalid
LICENSES/invalid-license-text|| bad license: invalid-license-text
LICENSES/Nokia-Qt-exception-1.1.txt|| deprecated license
LICENSES/MIT|| license without file extension
/tmp/pytest-of-nico/pytest-44/fake_repository14/invalid-license.py|| missing license invalid
LICENSES/MIT|| unused license
LICENSES/Nokia-Qt-exception-1.1.txt|| unused license
LICENSES/invalid-license-text|| unused license
/tmp/pytest-of-nico/pytest-44/fake_repository14/restricted.py|| read error
|| /tmp/pytest-of-nico/pytest-44/fake_repository14/file with spaces.py: without license
/tmp/pytest-of-nico/pytest-44/fake_repository14/file:with:colons.py|| without license
/tmp/pytest-of-nico/pytest-44/fake_repository14/no-license.py|| without license
|| /tmp/pytest-of-nico/pytest-44/fake_repository14/file with spaces.py: without copyright
/tmp/pytest-of-nico/pytest-44/fake_repository14/file:with:colons.py|| without copyright
/tmp/pytest-of-nico/pytest-44/fake_repository14/invalid-license.py|| without copyright
```

So probably **filenames will have to be escaped.**

nicorikken:
I originally had a test with a filename with colons, but this was causing errors on Windows. I removed it.

nicorikken:
The issue of using files with spaces seems to be an issue with the `errorformat` library: https://github.com/reviewdog/errorformat/issues/97 I don't think reuse-tool needs to escape filename output as there is already a clear separator in this output.
This leaves the edge-case of filenames with colons, but this is also rare in a development environment that will have to cater to Windows which doesn't allow colons in filenames.

src/reuse/lint.py
```diff
@@ -257,6 +264,64 @@ def custom_serializer(obj: Any) -> Any:
...
+
+    def license_path(lic: str) -> Optional[Path]:
+        "Resolve a license identifier to a license path."
+        return report.licenses.get(lic)
+
+    if not report.is_compliant:
+        # Bad licenses
+        for lic, files in sorted(report.bad_licenses.items()):
+            for file in sorted(files):
+                output.write(_(f"{file}: bad license {lic}\n"))
```
carmenbianca, line 288: Here and elsewhere, `_(f"{value}")` does not correctly translate. The formatting happens before the translation step here, while it should happen after.

carmenbianca (changes requested):
Thanks a lot for this feature @nicorikken ! Two very small nitpicks which I will patch up.

carmenbianca:
I also spent a while looking at escaping options. Thank you @nicorikken for your analysis. I think there are only two characters that really concern us here:

- Newlines in filenames (yes...)
- Colons in filenames

I gave it a spin with Pylint to see how other tools handle it. Here's the result:

```
$ pylint src/reuse
************* Module reuse.hello
world
src/reuse/hello
world.py:1:0: C0103: Module name "hello
world" doesn't conform to snake_case naming style (invalid-name)
************* Module reuse.hello:world
src/reuse/hello:world.py:1:0: C0103: Module name "hello:world" doesn't conform to snake_case naming style (invalid-name)
```

The answer is: don't.

I think we can be lazy about this and not bother.

carmenbianca (approved):

PR #960, opened 2024-04-18
Fix SPDX-License-Identifer typo.

BigBlueHat:


carmenbianca:
Good catch, thank you!

PR #968, opened 2024-04-27
Add support for any Java .properties

ForNeVeR:
Not sure if this has been discussed before, but all Java `.properties` files have a similar structure (and support Python-like comments), so it makes sense to support them all at once.


carmenbianca:
Seems fair to me. Thanks for your contribution @ForNeVeR. Can you add an entry in the change log?

ForNeVeR:
Yeah, sure, give me a minute :)

carmenbianca:
Fantastic, thank you very much, LGTM and merging :rocket:

PR #997, opened 2024-06-04
Add four new file(types) in the automatically supported files

waterkip:
* .taprc for node-tap
* .zshrc for zsh
* .t for perl test files
* .bats for BATS test files


carmenbianca:
Thanks @waterkip, merging this :+1:

PR #1008, opened 2024-06-18
Add support for ignoring files using REUSE.ignore

gergap:
I'm missing support to ignore certain files or folders for the license check in reuse-tool. In our repos (commercial product) there are certain files that will not be distributed, hence they don't need a license and should not appear in SPDX listing to avoid confusion. The existing solutions are not ideal:
* .gitignore: I don't want to ignore those files in Git, just for SPDX generation
* REUSE.toml: you can specify a license, but cannot ignore the file

I introduced a new file REUSE.ignore, which can be used to extend the IGNORE_FILE/DIR_PATTERNS in the code. This file is read at startup if it exists.

Example patterns:
```
^LICENSE-3RD-PARTY.TXT
^hooks/
^km/
^gitlab/
^src/test/afl/
````

Full regex syntax can be used. If the patterns ends with '/' it will be added to the `_IGNORE_DIR_PATTERNS` variable, otherwise `_IGNORE_FILE_PATTERNS` will be used. Therefor the trailing slash gets replaced with a '$' to terminate the expression and avoid wrong substring matches. More generic patterns are of course still possible using `^dir.*/` which will be expanded to `^dir.*$`.

### DIFF
--- a/src/reuse/__init__.py
+++ b/src/reuse/__init__.py
@@ -87,6 +87,7 @@ _IGNORE_SPDX_PATTERNS = [
 # Combine SPDX patterns into file patterns to ease default ignore usage
 _IGNORE_FILE_PATTERNS.extend(_IGNORE_SPDX_PATTERNS)
 
+
 def load_ignore_patterns() -> None:
     """Load ignore patterns from REUSE.ignore file."""
     ignore_file = "REUSE.ignore"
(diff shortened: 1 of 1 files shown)


carmenbianca:
Hi @gergap, and thanks for the contribution. Unfortunately, I have to close this. [This oft-requested feature is very specifically out-of-scope](https://reuse.software/faq/#exclude-file). And even if it weren't out-of-scope, the specification would need to be updated before the change is integrated, because the tool should reflect the specification.

For your case, you can use `.reuse/dep5` (or `REUSE.toml` when it is released) to achieve some of what you want. Set your company as copyright holder, and apply the license `LicenseRef-Proprietary` or some such, making sure to create the file `LICENSES/LicenseRef-Proprietary.txt`.

Alternatively, do your linting and/or SBOM generation on a filtered repository. Create some workflow that deletes the files which will not be distributed, and then run the linting/SBOM generation on the remaining subset of files.

I hope that this is moderately helpful.

Issue #604, opened 2022-10-12
Add filter for matlab files


jonasstein:
Matlab files (source code ends with `.m` comment is `%`) are not covered by reuse yet.

Matlab interprets the first lines of the code and the following comments in a preview function to help the to see what the function is good for.
If we add the license information in the first lines of the comment, the important preview renders unusable.
How should we use `SPDX` with matlab files, as the first comment lines after function are reserved for short documentation?

https://de.mathworks.com/help/matlab/matlab_prog/add-help-for-your-program.html

Please add a filter to reuse, so we can use it easily.

nicorikken:
Thanks for that insight @jonasstein There was already an issue open on supporting Matlab/Octave files #357 but that issue didn't speak of the preview.

To clarify, is this also true for comments on the top of the file? The documentation you linked shows an example of documentation below a function, but not at the top of a file.

One way to deal with this is to check if a comment block is present at the top of the file, and if so place SPDX headers below it. Would that work? Something like:

```matlab
% This script does awesome stuff.
% Pass in the necessary parameters to make it work.
% This section is intended for preview and is skipped over by REUSE tool.

% SPDX-FileCopyrightText: 2022 Person <person@server.org>
%
% SPDX-License-Identifier: GPL-3.0-or-later

function answer = important_question()
% This is function documentation, REUSE doesn't care about this.
answer = 42
```

Thanks to the linked documentation I was able to confirm the comment style conclusion from the other issue: https://de.mathworks.com/help/matlab/matlab_prog/comments.html

jonasstein:
@nicorikken Yes. There is a fruitful suggestion from the matlab staff on
https://de.mathworks.com/matlabcentral/answers/1840853-how-to-use-spdx-licenseinformation-in-mat-matlab-files

Issue #357, opened 2021-05-12
MATLAB/Octave comments


demsarjure:
Hi!

I just used the reuse-tool to license our code properly. The tool is great and I had no major issues when sorting things out. Thanks for preparing and maintaining the tool!

I would only recommend that you add the option of adding headers in MATLAB/Octave files (.m files). MATLAB/Octave are quite popular amongst the scientific community. I had to add headers in our code manually. I saw later from your source code that I could also use the option of adding headers to Tex files (LaTeX and MATLAB/Octave have same style of single line comments). In MATLAB/Octave single line comments start with "%", while block comments are encapsulated in 

%{ 
...
comments block
...
%}

I think this simple addition would make your tool even better and useful for a wider audience.

Cheers, Jure

nicorikken:
I started work in this, only to found out this is a bit more complicated. Both the `#` and `%` (and thus `#{` `#}` and `%{` `%}` are used for comments. From my history with Matlab and Octave I remember mainly using the `%`, but I did find the `#` syntax on GitHub as well. Some Octave documentation on the available comment syntax: https://octave.org/doc/v5.2.0/Comments.html

I'm not sure how to deal with 2 valid comment syntax in Reuse at the moment. There is a similar thing going on for the `BatchFileCommentStyle` for which multiple comment styles are valid. Thinking about how I would like to specify it, I would define multiple comment styles, and add them as a list to the `EXTENSION_COMMENT_STYLE_MAP` dictionary for a particular file extension. And the parser should then be able to select the applicable format and chose that over the others.

mxmehl:
@nicorikken Good question! Please feel free to make a suggestions. For *adding* a comment, I think we could stick to one syntax. Of course, one could try to find out an already used comment syntax but that seems to be a bit overload IMHO.

For *detecting* used comment styles, we may only have to add `%` if no comment style already uses that for single-line comments. `#` is already used by PythonCommentStyle so the linter should find it, right?

JoKeyser:
In Matlab, `#` is not recognized as a comment, so `%` should be used to generate headers.
For generation, I would also prefer single-line comments over multi-line comments.

For detection, it would of course be nice if all valid styles are recognized.

nicorikken:
Thanks @JoKeyser for sharing your insight. Then it should be easy to add support.

carmenbianca:
Closed by #1015

PR #1015, opened 2024-06-29
add octave/matlab file format support #357 #604 

kbroch-rivosinc:


carmenbianca:
Thanks @kbroch-rivosinc ! Merging.

Issue #993, opened 2024-05-28
Switch to importlib mode for pytest


carmenbianca:
Raised in #987

https://docs.pytest.org/en/stable/explanation/goodpractices.html#which-import-mode

PR #1046, opened 2024-07-08
Use importlib mode for pytest

khardix:
Attempt at implementing #993.

- Treat `tests/` as namespace and relative-import `conftest`. 
- Switch the mode in `pyproject.toml`.

      changed files.

### DIFF
--- a/changelog.d/changed/importlib_mode_pytest.md
+++ b/changelog.d/changed/importlib_mode_pytest.md
@@ -1,0 +1,3 @@
+- Tests now use
+  [`--import-mode=importlib`](https://docs.pytest.org/en/stable/explanation/goodpractices.html#which-import-mode).
+  (#1046)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,7 +148,7 @@ use_parentheses = true
 line_length = 80
 
 [tool.pytest.ini_options]
-addopts = "--doctest-modules"
+addopts = "--doctest-modules --import-mode=importlib"
 
 [tool.mypy]
 files = [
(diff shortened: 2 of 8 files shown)


khardix:
Rebased on the current `main`. I was unable to convince mypy to follow the now-relative `.conftest` import without turning `tests/` into a proper python package, so there is a new empty `__init__.py` file in there. I hope this is not an issue. :crossed_fingers:

khardix:
@floriansnow Ping, this should be ready to review.

carmenbianca:
Hi @khardix, thanks for the PR.

I will close this because it doesn't really seem like an improvement over the status quo, and the [relevant pytest web page](https://docs.pytest.org/en/latest/explanation/pythonpath.html) now says:

> Initially we intended to make importlib the default in future releases, however it is clear now that it has its own set of drawbacks so the default will remain prepend for the foreseeable future.

And importantly, under 'disadvantages', it lists this:

> Testing utility modules in the tests directories (for example a tests.helpers module containing test-related functions/classes) are not importable. The recommendation in this case is to place testing utility modules together with the application/library code, for example app.testing.helpers.

Because we do have 'test utility [functions]' in `conftest.py`, we need to be able to import them. Moving this stuff to the module being tested seems a little silly to me.

Adding `__init__.py` to the `tests/` directory remains disrecommended by pytest, so that's sadly a non-starter.

khardix:
Acknowledged. I was proposing this change because the `prepend` method was failing when building the Fedora rpm (for some reason), and with the `importlib` method being the future™ at that point, it seemed to be the best solution moving forwards. With the rescinding of that proposal from the pytest side, I'll go back to debugging why it fails for me.

PR #1052, opened 2024-07-19
Add support for GitHub annotations in GitHub actions

nijel:
- [x] Wrote tests.
      changed files.


carmenbianca:
Wow, amazing PR @nijel ! Thank you for this work.

I wonder whether it makes sense to keep `--lines` in parallel to `--github`. They do effectively the same. For backward compatibility, we should probably keep both, though…

Do you know whether this GitHub format of errors is also consumed by other programs?

nijel:
`--lines` is still more user friendly, so I'd keep both. I think gitlab has a similar format, but I have no experience with that. https://docs.astral.sh/ruff/settings/#output-format  supports that in addition to GitHub...

src/reuse/lint.py
```diff
@@ -264,7 +267,9 @@ def custom_serializer(obj: Any) -> Any:
     )
 
 
-def format_lines(report: ProjectReport) -> str:
+def get_errors(
+    report: ProjectReport,
+) -> Iterable[tuple[Path | str | None, str]]:
```
carmenbianca, line 279: The tuple should probably be a NamedTuple

carmenbianca: Also, I don't think `Iterable` is correct. Generators in Python are annoying to type correctly. I think it's `Iterator[type, False, False]` or some such. I'll look this up.

src/reuse/lint.py
```diff
@@ -270,15 +276,15 @@ def custom_serializer(obj: Any) -> Any:
 def get_errors(
     report: ProjectReport,
 ) -> Iterable[tuple[Path | str | None, str]]:
-    """Formats data dictionary as plaintext strings to be printed to sys.stdout
+    """Returns data dictionary iterable of paths and errors.
```
carmenbianca: This docstring needs a little workshopping.

src/reuse/lint.py
```diff
@@ -343,6 +349,27 @@ def format_lines(report: ProjectReport) -> str:
     return ""
 
 
+def format_github(report: ProjectReport) -> str:
+    """Formats data dictionary as GitHub workflow commands
+    to be printed to sys.stdout
```
carmenbianca, line 355: A link to the documentation would be nice.

docs/man/reuse-lint.rst
```diff
@@ -103,3 +103,12 @@ Options
...
   Display help and exit.
+
+Environment
+-----------
+
+.. envvar:: REUSE_OUTPUT_FORMAT
+
+   Specifies output format, one of ``plain``, ``lines``, ``github``, ``json``
+
+   It behaves same as corresponding command line options.
```
carmenbianca, line 114: There should be a documented order of precedence here. If both `REUSE_OUTPUT_FORMAT` and a CLI flag are used, which takes precedence?

As written in the code, I believe the env var takes precedence. I do not know if this is desirable behaviour.

carmenbianca: Additionally, should `quiet` be a valid value here?

carmenbianca: Additionally, what should happen if the value of `REUSE_OUTPUT_FORMAT` is invalid?

carmenbianca (changes requested):
Some questions and feedback. The PR looks great overall.

I will write a few patch commits to fix the easy stuff that doesn't need discussion.

docs/man/reuse-lint.rst
nijel: Ah, seems like I didn't notice this feedback, sorry.

With the current implementation, the environment variable takes precedence. This is intentional to allow having command line configuration and override it by environment if executed in environment which might want different output (as GitHub).

Unknown values are simply discarded, but it might be a better approach to issue a warning.

ferdnyc: @nijel 

> having command line configuration and override it by environment if executed in environment which might want different output (as GitHub).

That's an argument I've seen before, in other contexts, and it's always felt upside-down to me.

Typically the precedence goes config file -> envvar -> commmand-line flags. You can store your configuration in a file, but if need be can still override it by setting variables in the environment. Above all else, though, the arguments specified directly on the command line supersede any other sources.

It's a Principle of Least Astonishment thing. Using examples from some common build systems, if I run a command like,

```bash
$ cmake -B _build -S . -DCMAKE_INSTALL_PREFIX=/tmp
```

then for it to install somewhere else because I happened to have `CMAKE_INSTALL_PREFIX="/var/tmp"` set in the environment, then that's a very astonishing invisible override. 

By the same token, it was [a bug](https://github.com/rust-lang/cargo/pull/11077) in Rust's Cargo a while back that a command like:

```bash
$ CARGO_BUILD_TARGET_DIR='env' cargo c --config myConfig.toml
```

...Would ignore a target dir specified in `myConfig.toml`. So, that was fixed. 

BUT, that's actually another example of the previous form (command-line flags overriding the environment), because the config is loaded explicitly via the command-line args. Cargo has always, correctly, allowed a command like this:

```bash
$ CARGO_BUILD_TARGET_DIR='env' cargo c
```

...to override the target dir set in its _default_, autoloaded `.cargo/config.toml` file.

Seems to me that. if you want to set values via the environment, _**don't**_ put them on the command line. Or, use the envvars on the command line, possibly with defaults. 

For example:

```bash
make ${BUILD_TARGET:-all}
```

...will run `make all` unless `${BUILD_TARGET}` is set in the environment.

But a `make all` should never ignore that explicit `all` and build something _different_, just because there's some value set in a config file or an environment variable. IMHO.

nijel: You're right, the precedence should be different. I thought about being able to override it when executed in GitHub actions, but reuse is there executed without a command line argument, so the environment would be used even if it is not a priority.

I see @carmenbianca has now made some changes on top of mine, so I'm not going to touch the PR unless it is desired to not interfere with other changes.

Issue #1057, opened 2024-08-07
REUSE.toml data seems to be unintentionally ignored if a .license file is found


nea89o:
If a `.license` file only contains copyright or only contains licensing information the other information may be provided by a `REUSE.toml` file (at least to my understanding of the spec).

<details>
<summary>
Example configuration: 
</summary>

```bash
$ cat circle.png.license
SPDX-FileContributor: June Hibiscus
$ cat REUSE.toml
#SPDX-FileCopyrightText: 2024 Linnea Gräf <nea@nea.moe>
#
#SPDX-License-Identifier: CC0-1.0

version = 1
SPDX-PackageName = "Firmament"
SPDX-PackageSupplier = "Linnea Gräf <nea@nea.moe>"
SPDX-PackageDownloadLocation = "https://github.com/nea89o/Firmament"

[[annotations]]
path = "**/*.png"
SPDX-License-Identifier = "CC-BY-4.0"
SPDX-FileCopyrightText = "Linnea Gräf <nea@nea.moe>"
$ reuse lint
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
* /home/nea/test/Firmament/circle.png


# SUMMARY

* Bad licenses: 0
* Deprecated licenses: 0
* Licenses without file extension: 0
* Missing licenses: 0
* Unused licenses: 0
* Used licenses: CC0-1.0
* Read errors: 0
* Files with copyright information: 1 / 2
* Files with license information: 1 / 2

Unfortunately, your project is not compliant with version 3.2 of the REUSE Specification :-(


# RECOMMENDATIONS

* Fix missing copyright/licensing information: For one or more files, the tool
  cannot find copyright and/or licensing information. You typically do this by
  adding 'SPDX-FileCopyrightText' and 'SPDX-License-Identifier' tags to each
  file. The tutorial explains additional ways to do this:
  <https://reuse.software/tutorial/>
```
</details>

## There are some relevant pieces in the code here:

https://github.com/fsfe/reuse-tool/blob/37f342d96b11419516ecc3d7c25fac0fd916c727/src/reuse/project.py#L292-L307

This piece of code is responsible for copying over "closest" `REUSE.toml` data if only the licensing info or only the copyright info is present in the `file_result` (which comes from the `.license` file).

https://github.com/fsfe/reuse-tool/blob/37f342d96b11419516ecc3d7c25fac0fd916c727/src/reuse/project.py#L253-L257

[...]

PR #1058, opened 2024-08-07
Fix global licensing being ignored with a .license file

nea89o:
Fixes https://github.com/fsfe/reuse-tool/issues/1057

- [x] Wrote tests.
- [ ] Documented my changes in `docs/man/` or elsewhere. (n/a)
      changed files.

### DIFF
--- a/src/reuse/project.py
+++ b/src/reuse/project.py
@@ -4,6 +4,7 @@
 # SPDX-FileCopyrightText: 2023 Matthias Riße
 # SPDX-FileCopyrightText: 2023 DB Systel GmbH
 # SPDX-FileCopyrightText: 2024 Kerry McAdams <github@klmcadams>
+# SPDX-FileCopyrightText: 2024 Linnea Gräf
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -243,7 +244,7 @@ class Project:
 
         # Search the global licensing file for REUSE information.
         if self.global_licensing:
-            relpath = self.relative_from_root(path)
+            relpath = self.relative_from_root(original_path)
             global_results = defaultdict(
                 list, self.global_licensing.reuse_info_of(relpath)
             )
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -457,12 +457,10 @@ def test_reuse_info_of_copyright_xor_licensing(empty_directory):
     assert not bar_file_info.spdx_expressions
 
 
-def test_reuse_info_of_copyright_xor_licensing_with_license_dot(
-    empty_directory,
-):
-    """Test a corner case where partial REUSE information is defined inside of a
-    .license file (copyright xor licensing). Get the missing information from
-    the REUSE.toml.
+def test_reuse_info_of_reuse_toml_dot_license(empty_directory):
+    """Test a corner case where there is REUSE information inside of a file, its
+    .license file, and REUSE.toml. Only the REUSE information from the .license
+    file and REUSE.toml should be applied to this file.
     """
     (empty_directory / "REUSE.toml").write_text(
         cleandoc(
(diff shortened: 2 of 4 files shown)


src/reuse/project.py
```diff
@@ -251,7 +252,7 @@ def reuse_info_of(self, path: StrPath) -> List[ReuseInfo]:
 
         # Search the global licensing file for REUSE information.
         if self.global_licensing:
-            relpath = self.relative_from_root(path)
+            relpath = self.relative_from_root(original_path)
```
carmenbianca, line 255: I think this contribution is correct.

tests/test_project.py
```diff
@@ -647,6 +648,80 @@ def test_reuse_info_of_copyright_xor_licensing(empty_directory):
...
+
+            [[annotations]]
+            path = "*.py"
+            SPDX-FileCopyrightText = "2017 Jane Doe"
+            SPDX-License-Identifier = "CC0-1.0"
+
+            [[annotations]]
+            path = "*.py.license"
+            SPDX-FileCopyrightText = "2017 Jane Doe"
+            SPDX-License-Identifier = "MIT"
```
carmenbianca, line 671: This is a little strange to me, and seems incorrect. A `.license` file cannot/does not have licensing itself. The logic of REUSE is that these `.license` files are direct substitutes licensing-wise. This means that `foo.py.license` completely overrides the REUSE information inside of `foo.py`, and whenever `foo.py` is referenced anywhere (in REUSE.toml or in the code), we should really be looking at the contents of `foo.py.license` instead.

The second `[[annotations]]` table should probably be removed.

tests/test_project.py
```diff
@@ -647,6 +648,80 @@ def test_reuse_info_of_copyright_xor_licensing(empty_directory):
...
+        )
+    )
+    project = Project.from_directory(empty_directory)
+
+    foo_infos = project.reuse_info_of("foo.py")
+    assert len(foo_infos) == 2
+    foo_toml_info = [info for info in foo_infos if info.spdx_expressions][0]
+    assert foo_toml_info.source_type == SourceType.REUSE_TOML
+    assert not foo_toml_info.copyright_lines
+    assert "MIT" not in str(foo_toml_info.spdx_expressions)
```
carmenbianca, line 710:
```suggestion
    assert "CC0-1.0" in str(foo_toml_info.spdx_expressions)
```

I think it's better to be explicit here.

tests/test_project.py
```diff
@@ -647,6 +648,80 @@ def test_reuse_info_of_copyright_xor_licensing(empty_directory):
...
+    foo_file_info = [info for info in foo_infos if info.copyright_lines][0]
+    assert foo_file_info.source_type == SourceType.DOT_LICENSE
+    assert not foo_file_info.spdx_expressions
```
carmenbianca, line 713:
```suggestion
    assert foo_file_info.source_type == SourceType.DOT_LICENSE
    assert not foo_file_info.spdx_expressions
    assert "SPDX-FileCopyrightText: 2017 John Doe" in foo_file_info.copyright_lines
```

Explicit again to make the test more robust.

carmenbianca (changes requested):
Hi @nea89o and thanks for your contribution! I struggled a bit with the reading of the issue, so I'm going to re-summarise to make sure I understand correctly:

Using `closest` logic, if there is copyright information inside of a file, and licensing information inside of `REUSE.toml`, then they are both applied to the file.

However, as I understand it, the tool currently poorly/incorrectly handles the situation where the copyright information is inside of the file's `.license` file.

Does that seem right?

Thanks again for your contribution!

tests/test_project.py
nea89o: The old behaviour would look up the information for the .license file path in the REUSE.toml instead of looking up the non .license file path. Because of that I thought i would add an explicit test that the .license file information in the REUSE.toml is ignored.

carmenbianca: I had another look at this. It took me a while to get it, but I think I understand. You're testing here that `.license` files are not valid targets inside of a `REUSE.toml` files, and should therefore be ignored.

I've moved this to a different test, to reduce the amount of stuff happening inside of one test. I've also simplified the initial test a little by reducing the amount of files, and by using `aggregate` instead of `closest`.

Thanks for your contribution!

carmenbianca (approved):
LGTM after a little patching. Thanks @nea89o !

PR #1061, opened 2024-08-13
Add `.envrc` and `flake.lock` extensions

bpeetz:
I've added the `.envrc` (used by at least [direnv](https://direnv.net/)) and the nix flake lock file (`flake.lock`) to the recognized extensions. I've not added an entry in the changelog, as it seemed to tiny to be included in it (let me know if I should include it.). 

      changed files.


carmenbianca:
Thanks @bpeetz !

PR #1089, opened 2024-10-14
Add `.cabal` and `cabal.project` as recognised file types for comments

ozkutuk:
This PR adds `.cabal` and `cabal.project` as recognized file types for comments. These files are used by the [Cabal](https://www.haskell.org/cabal/) build tool.

      changed files.

### DIFF
--- a/changelog.d/added/comment-cabal.md
+++ b/changelog.d/added/comment-cabal.md
@@ -1,0 +1,2 @@
+- Added `.cabal` and `cabal.project` (Haskell) as recognised file types for
+  comments. (#1089)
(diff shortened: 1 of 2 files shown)


ozkutuk:
Hmm, I have now realized that the modified `.cabal` files are not accepted by Cabal anymore, because Cabal mandates that the `cabal-version` field is the first thing that appears in the file, apparently including comments:
```
Errors encountered when parsing cabal file ./test.cabal:

test.cabal:6:1: error:
cabal-version should be at the beginning of the file starting with spec version 2.2.
Current cabal-version values are listed at https://cabal.readthedocs.io/en/stable/file-format-changelog.html.

    6 | cabal-version:      3.0
      | ^
```

There is nothing wrong with this PR per se, but this will probably not be of much use unless this is somehow resolved on the Cabal side.

carmenbianca:
Oops, I hadn't read your second comment! That is indeed a bit of a thing. You could solve that by adding `cabal-version:` as a shebang to the HaskellCommentStyle class. It's not 100% perfect, but should work.

PR #1117, opened 2024-12-20
feat(BEAM): Add support for Erlang, Elixir, and Gleam comments

kikofernandez:
Adds support for comments for the most used BEAM languages:
- Erlang
- Elixir
- Gleam

- [X] Added a change log entry in `changelog.d/<directory>/`.
- [X] Added self to copyright blurb of touched files.
- [X] Added self to `AUTHORS.rst`.
- [X] My changes do not contradict
      changed files.

### DIFF
--- a/src/reuse/comment.py
+++ b/src/reuse/comment.py
@@ -404,6 +404,15 @@ class ErlangCommentStyle(CommentStyle):
 
 
 class ElixirCommentStyle(CommentStyle):
+    """Elixir comment style."""
+
+    SHORTHAND = "elixir"
+
+    SINGLE_LINE = "#"
+    INDENT_AFTER_SINGLE = " "
+    SHEBANGS = ["#!"]
+
+
 class FtlCommentStyle(CommentStyle):
     """FreeMarker Template Language comment style."""
 
@@ -644,10 +653,10 @@ EXTENSION_COMMENT_STYLE_MAP = {
     ".dts": CppCommentStyle,
     ".dtsi": CppCommentStyle,
     ".el": LispCommentStyle,
-    ".ex": PythonCommentStyle,
-    ".exs": PythonCommentStyle,
     ".erl": ErlangCommentStyle,
     ".escript": ErlangCommentStyle,
+    ".ex": ElixirCommentStyle,
+    ".exs": ElixirCommentStyle,
     ".f": FortranCommentStyle,
     ".fsproj": HtmlCommentStyle,
     ".f03": ModernFortranCommentStyle,
--- a/src/reuse/comment.py
+++ b/src/reuse/comment.py
@@ -19,6 +19,7 @@
 # SPDX-FileCopyrightText: 2024 Rivos Inc.
 # SPDX-FileCopyrightText: 2024 Anthony Loiseau <anthony.loiseau@allcircuits.com>
 # SPDX-FileCopyrightText: 2025 Raphael Schlarb <info@raphael.schlarb.one>
+# SPDX-FileCopyrightText: 2025 Kiko Fernandez-Reyes <kiko@erlang.org>
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
--- a/src/reuse/comment.py
+++ b/src/reuse/comment.py
@@ -358,6 +358,7 @@ class CppSingleCommentStyle(CommentStyle):
 
     SINGLE_LINE = "//"
     INDENT_AFTER_SINGLE = " "
+    SHEBANGS = ["#!"]  # Gleam
 
 
 class EmptyCommentStyle(CommentStyle):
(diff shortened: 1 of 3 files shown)


kikofernandez:
Thanks @maennchen for reminding me of Erlang's lexer and parser generators, which are now included in the PR :)

src/reuse/comment.py
```diff
@@ -392,6 +393,26 @@ class ModernFortranCommentStyle(CommentStyle):
...
+    """Erlang comment style."""
+
+    SHORTHAND = "erlang"
+
+    SINGLE_LINE = "%"
+    INDENT_AFTER_SINGLE = " "
+    SHEBANGS = ["#!"]
+
+
+class ElixirCommentStyle(CommentStyle):
```
floriansnow, line 406: This looks identical to `PythonCommentStyle`. Do you see any differences or do you think they styles will diverge in the future?

src/reuse/comment.py
```diff
@@ -400,6 +421,16 @@ class FtlCommentStyle(CommentStyle):
     MULTI_LINE = MultiLineSegments("<#--", "", "-->")
 
 
+class GleamCommentStyle(CommentStyle):
```
floriansnow, line 424: I think `CppCommentStyle` will serve the same purpose here, won't it?

src/reuse/comment.py
kikofernandez: I do not see any differences, but I thought that maybe one does not want to conflate comment styles for different languages, in case in a few years one comment style changes a bit, while the other should not.

Happy to be wrong and I can fix to use Python style, if you think this is more maintainable 😃

src/reuse/comment.py
kikofernandez: I am not sure it would work, as Gleam does not have multiline as CPP
```
-- copy-pasted CppCommentStyle snippet
MULTI_LINE = MultiLineSegments("/*", "*", "*/")
```
and the CppCommentStyle also uses some php SHEBANG which I do not think work on Gleam
```
-- copy-pasted CppCommentStyle snippet
SHEBANGS = [
        "#!",  #  V-Lang
        "<?php",  # PHP
    ]
```

floriansnow: If I'm not completely off, using `CppCommentStyle` would only cause a problem if there was a chance for something that looks like a multiline comment in Gleam, but isn't one. The same should also be true for the shebang. @carmenbianca can you confirm?

src/reuse/comment.py
floriansnow: I mean I'm just getting back into the codebase, but from what I can see, many languages use the same code style definition and it wouldn't be an issue to add another one should they diverge in the future.

src/reuse/comment.py
kikofernandez: To be honest, I do not think it is an issue in Gleam, but I have not written much Gleam myself.
I mention this as to separate concerns :) 
Please let me know what the preferred way would be :)

src/reuse/comment.py
kikofernandez: Ok, can you confirm that I understand that the preference is to follow `PythonCommentStyle`? 
Just to confirm that I understood correctly that I should modify this

floriansnow: I indeed did not miss anything here and for identical comment styles, we use an existing one. So in this case, `PythonCommentStyle` is identical and we can go with that.

src/reuse/comment.py
floriansnow: The shebangs should not create any problems (I think it's unlikely you'll find `<?php` in Gleam code) as REUSE will never add them, simply change the position of the added header in case it encounters as shebang in the first line of the file.

The multiline comments would create an issue however, if REUSE was forced to create a multiline header. So I think `CppSingleCommentStyle` would work, though. We can add the shebang for Gleam together with a comment explaining what it's for (just like in the example you pasted above).

kikofernandez: I understood this comment as create a subtype of `CppSingleCommentStyle` and add the `SHEBANG` that Gleam uses. I also added comments, but I can use more guidance if this is not correct.

Thanks a lot for taking the time to review this PR.

floriansnow: I see. That's close to what I meant: I meant just using `CppSingleCommentStyle` directly and expanding it with the Gleam shebang. Sorry for the confusion.

kikofernandez: I hope I got it right this time around :) 
Thanks again for the feedback, and sorry for the mistakes, Python is not my strong suite.

kikofernandez:
Is there anything else that I can help with to get support for Erlang/Elixir/Gleam?
Thanks!

kikofernandez:
Thanks @floriansnow for your guidance

PR #1135, opened 2025-01-28
Add helper script snippet for following renames

mschoettle:
The current example does not follow renames which I think should be included to know the very first time the file was introduced.
For some reason, `git log --reverse --follow` does not work properly. So reversed it to use `tail` instead of `head`.

I added an extra section but it might be worth to add it to the existing examples instead.

      changed files.

>  - [ ] Added a change log entry in `changelog.d/<directory>/`.

I feel that the changes are too small and do not affect the tool itself, so I did not add one. If you disagree, I'll be happy to add a changelog entry.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -145,3 +145,4 @@ Contributors
 - Emil Velikov <emil.l.velikov@gmail.com>
 - Linnea Gräf <nea@nea.moe>
 - Raphael Schlarb <info@raphael.schlarb.one>
+- Matthias Schoettle <opensource@mattsch.com>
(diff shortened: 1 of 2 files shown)


mschoettle:
@floriansnow @carmenbianca Can one of you please review?

floriansnow:
Thank you for your contribution! This looks good and useful to me! I agree that this does not warrant an entry in the changelog, but please add your copyright notice in the changed file. We want to track authors properly, even if it's a small change.

mschoettle:
@floriansnow Thanks for the feedback. Added the copyright notice to the file.

PR #1136, opened 2025-01-30
Skip newline before python docstring

charlie-rivos:
Docstrings are defined in PEP 257 as "A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition." When reuse adds a header, it forcibly adds a space before the docstring violating this principle.

When the style is set to Python and the first line in the file is a docstring, do not add a space between the license and the docstring.

More information can be seen here: https://peps.python.org/pep-0257/

- [ ] Added self to copyright blurb of touched files.
- [ ] Added self to `AUTHORS.rst`.
- [ ] Wrote tests.
- [ ] My changes do not contradict
      changed files.


carmenbianca:
Hi @charlie-rivos, sorry for the late response. I discussed this with @floriansnow, and the PR would be great if it were generalised!

That is to say, given `foo.py`:

```python
# Copyright Jane Doe
pass
```

Running `reuse annotate --license EUPL-1.2` should produce:

```python
# Copyright Jane Doe
#
# SPDX-License-Identifier: EUPL-1.2
pass
```

instead of:

```python
# Copyright Jane Doe
#
# SPDX-License-Identifier: EUPL-1.2

pass
```

The word 'pass' is irrelevant here. It could be _any_ code, in _any_ language. Put in different words: If there is no empty line between the top header and the code, `reuse annotate` should not add one.

(However, if there is no top header, `reuse annotate` should add an empty line as a stylistic preference.)

charlie-rivos:
Thank you for taking the time to review this! I updated the MR to be generic and not just for Python. I had to update the respective test cases.

I mention this in the commit description as well -- I feel like it is valid to include the newline immediately after the license if there would otherwise be a comment of the same comment style there. Please let me know if you agree!

floriansnow:
Thank you for generalizing this!

> I feel like it is valid to include the newline immediately after the license if there would otherwise be a comment of the same comment style there.

Yes, but also in the case that Carmen mentioned where there is just code afterwards. It's a stylistic choice and when placing the header, I think it would be good to add an empty line after. If there's already a license header there, we shouldn't touch either case.

charlie-rivos:
> Yes, but also in the case that Carmen mentioned where there is just code afterwards. It's a stylistic choice and when placing the header, I think it would be good to add an empty line after. If there's already a license header there, we shouldn't touch either case.

Hmm okay, that seems reasonable. I can change it to work like that.

changelog.d/changed/persist-whitespace.md
```diff
@@ -0,0 +1,2 @@
+- To avoid improperly formatting files, place license without impacting
+  whitespace. (#1136)
```
carmenbianca, line 2: This should be clearer for a reader who doesn't know what is going on.

src/reuse/header.py
```diff
@@ -205,16 +206,35 @@ def _extract_shebang(prefix: str, text: str) -> tuple[str, str]:
...
             break
-    shebang = "\n".join(shebang_lines)
+    shebang = "".join(shebang_lines)
     return (shebang, text)
 
 
+def place_header(
+    header: str, before: str, after: str, has_existing_header: bool
+) -> str:
+    """Combine the header and the text from the file into the final output"""
```
carmenbianca, line 222: 'From the file' is incorrect; this function deals with strings.

src/reuse/header.py
```diff
@@ -205,16 +206,35 @@ def _extract_shebang(prefix: str, text: str) -> tuple[str, str]:
...
+    new_text = f"{header}\n"
+    if before.strip():
+        new_text = f"{before.rstrip()}\n\n{new_text}"
+    if after.strip():
+        # Create space between header and following code only if a newline
+        # doesn't already exist, and there wasn't previously a header.
+        if not (has_existing_header or after.startswith("\n")):
```
carmenbianca, line 229:
```suggestion
        if not has_existing_header and not after.startswith("\n"):
```

This is more easily parsed by humans.

src/reuse/header.py
```diff
@@ -205,16 +206,35 @@ def _extract_shebang(prefix: str, text: str) -> tuple[str, str]:
...
+def place_header(
+    header: str, before: str, after: str, has_existing_header: bool
+) -> str:
+    """Combine the header and the text from the file into the final output"""
+    new_text = f"{header}\n"
+    if before.strip():
+        new_text = f"{before.rstrip()}\n\n{new_text}"
+    if after.strip():
+        # Create space between header and following code only if a newline
+        # doesn't already exist, and there wasn't previously a header.
```
carmenbianca, line 228: There should be a test for the scenario where there is a file WITHOUT header, but WITH newlines at the start.

src/reuse/header.py
```diff
@@ -253,6 +273,8 @@ def find_and_replace_header(
     except MissingReuseInfoError:
         before, header, after = "", "", text
 
+    has_existing_header = bool(header)
+
```
carmenbianca, line 277: This is not needed as a variable. Instead, we shouldn't re-define the variable `header` further down in this function. We should define a new variable `result` or something.

tests/test_header.py
```diff
@@ -244,6 +245,21 @@ def test_find_and_replace_verbatim():
     assert find_and_replace_header(text, info) == text
 
 
+def test_find_and_replace_verbatim_no_newline():
+    """Replace a header with itself with no newline."""
```
carmenbianca, line 249: This docstring would benefit from being a little clearer.

```suggestion
    """Do not add an empty line after existing headers."""
```

carmenbianca (changes requested):
@floriansnow We did this review on call. The notes :)

Thank you @charlie-rivos for your PR! Do you want to follow up on this? If not, we can fix up the tiny things and merge.

src/reuse/header.py
charlie-rivos: I added two tests for that, one with one space, and one with multiple spaces

charlie-rivos:
> Thank you @charlie-rivos for your PR! Do you want to follow up on this? If not, we can fix up the tiny things and merge.

I have updated according to the suggestions, if you have any differing ideas on how I resolved the suggestions please feel free to make the appropriate changes!

PR #1145, opened 2025-03-04
Implement YearRange and CopyrightNotice

### LINKED ISSUE #947: No longer make copyright lines stringly typed
Presently, copyright lines are stringly typed in the tool. The whole thing `SPDX-FileCopyrightText: 2016, 2018--2020, 2023--present Carmen Bianca BAKKER <carmenbianca@fsfe.org>` is one string.

I think it might make sense to write some kind of `CopyrightLine` class that can parse elements from such strings. For example, the above properties of the above example:

- copyright_prefix - `SPDX-FileCopyrightText`
- date_range - some DateRange object that can encode the concept of `2016`, `2018--2020`, and `2023--present`. Or, if that all is too over-engineered, just the string `2016, 2018--2020, 2023--present`
- name - `Carmen Bianca BAKKER`
- contact - `carmenbianca@fsfe.org`

If implemented well, this could heavily simplify date-range merging, which is currently spaghetti code.

carmenbianca:
Fixes #947.

  - [x] Specifically, `reuse annotate --year` works differently now.
- [x] Wrote tests.
      changed files.


carmenbianca:
So this is basically done. All tests pass, and the whole code base is converted to using these new data types instead of strings. There are probably some comments or variable names that still hint at the old string types, or patterns in the code that make less sense given the new data types, but on the whole, everything appears to work.

The benefits of doing this aren't immediately obvious yet, except that #328 is now implemented in an obvious and elegant way that doesn't require dark string manipulation magics.

The downside is that the new data type is somewhat less flexible than strings. It has some assumptions (years come before the author, separated by spaces, dashes, and commas; contacts are between square brackets at the end; etc etc etc), and tries to remain flexible outside of those assumptions (that is: put everything that isn't recognised into the `name` field of `CopyrightNotice`), but it's not infeasible that a bug report will be opened by someone because of something I had not anticipated.

The other downside is that one test is broken on Windows. I'll fix that, probably.

PR #1150, opened 2025-03-14
cli/supported_licenses: simplify the license list

perillo:
Remove the license reference, since it may not fit the terminal.

Move the license name before the license id, in order to improve the user experience.

Make the text width dynamic, in order to improve readability.

- [ ] Added self to copyright blurb of touched files.
- [ ] Added self to `AUTHORS.rst`.
- [ ] Wrote tests.
- [ ] My changes do not contradict
      changed files.

### DIFF
--- a/src/reuse/cli/supported_licenses.py
+++ b/src/reuse/cli/supported_licenses.py
@@ -20,10 +20,15 @@ def supported_licenses() -> None:
     # pylint: disable=missing-function-docstring
     licenses = _load_license_list(_LICENSES)[1]
 
+    license_name_width = 0
     for license_id, license_info in licenses.items():
         license_name = license_info["name"]
-        license_reference = license_info["reference"]
-        click.echo(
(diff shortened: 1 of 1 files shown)


perillo:
I started to use REUSE and I noted that the `supported_licenses` command it not very user friendly.

Each lines are too long, and on sway + foot I'm unable to select the license URL.

So I decided to remove the license reference.

If you agree to this change, it should possible to add an option to show all the details of a single license, with all supported metadata like link, OSI approved and FSFLibre approved;
see https://github.com/spdx/license-list-data/blob/main/licenses.md as an examples.

It would be useful to add filters based on OSI and FSFLibre compliance.

Thanks.

floriansnow:
Thank you for this contribution. However, I'm not convinced this is an issue yet. Would you mind describing in more detail, perhaps with some screenshots, how that issue affects you?

PR #1151, opened 2025-03-14
Add Github action for closing stale issues

floriansnow:
This will help with keeping the repo in a healthy state. Note that while this does contain a message for marking PRs as stale and closing them, it doesn't currently do so because the timeframes for that are set to -1. I just added the text to make activating it as painless as possible if we ever decide to do so.

Also, this will only touch 5 issues per run, so the pings will get spread out nicely. We can increase the number once it's handled all issues once or we can leave it as is until we notice a bottleneck.


carmenbianca (approved):
LGTM

PR #1161, opened 2025-04-07
Deprecate OTP linking exception

kikofernandez:
Deprecates the Erlang/OTP linking exception. 

This exception refers to when Erlang/OTP had the Erlang Public License but Erlang/OTP has Apache 2.0 since 2016 or so. 
I thought that this exception should be considered deprecated, any recent and supported Erlang version (Erlang/OTP maintains the last three major releases) is not licensed under EPL.

If this exception should not be deprecated, please close this PR.

- [X] Added a change log entry in `changelog.d/<directory>/`.
- [X] My changes do not contradict
      changed files.

### DIFF
--- a/src/reuse/resources/exceptions.json
+++ b/src/reuse/resources/exceptions.json
@@ -192,7 +192,7 @@
     },
     {
       "reference": "./erlang-otp-linking-exception.json",
-      "isDeprecatedLicenseId": false,
+      "isDeprecatedLicenseId": true,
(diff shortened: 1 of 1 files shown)


carmenbianca:
Hi @kikofernandez 

We get this file from https://github.com/spdx/license-list-data

The correct place to make this PR is https://github.com/spdx/license-list-XML, I think.

I hope that helps. I will close this PR.

PR #1191, opened 2025-06-09
Correct invocation for jujutsu file listing

### LINKED ISSUE #1166: Jujutsu integration is broken starting with jujutsu 0.26.0
The jujutsu integration introduced in #711 [uses `jj files`](https://github.com/fsfe/reuse-tool/blob/c435d8c8520085bd4a2957b6623be6898f8b62bb/src/reuse/vcs.py#L264) to get the repository files. This command [was deprecated in jujutsu 0.19.0](https://github.com/jj-vcs/jj/blob/e5478bbf7b3b0aee4141a29d182fd501ae191070/CHANGELOG.md#deprecations-10), and [removed in jujutsu 0.26.0](https://github.com/jj-vcs/jj/blob/e5478bbf7b3b0aee4141a29d182fd501ae191070/CHANGELOG.md#breaking-changes-3). The new blessed way to retrieve a list of files is to use  `jj file list` instead.

I am willing to contribute the fix (i.e. call `jj file list` instead of `jj files`), but wanted to inquire about backwards compatibility first: the fix would break users using versions older than jujutsu 0.19.0, which was released in 2024-07-03. Should reuse check the jujutsu version and call the appropriate subcommand? Or is it fine to break the integration for users with older versions?

PigeonF:
The `jj files` command was deprecated in jujutsu 0.19.0, and removed in jujutsu 0.26.0 (released in february of 2025). The fix is to use the new `jj file list` command instead.

Before adding a note to the changelog I wanted to ask about backwards compatibility. This PR fixes use with jujutsu >= 0.26.0, but breaks users with jujutsu < 0.19.0 (which was released in july 2024). In theory the code in `vcs.py` could check which version of jujutsu is installed and then use the corresponding correct command, but I am not sure what the project policy is with regards to backwards compatibility vs maintainer burden, so the current fix breaks backwards compatibility with older versions in order to keep the code simple.

- [ ] Wrote tests.
      changed files.

Fixes #1166.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -150,3 +150,4 @@ Contributors
 - Kiko Fernandez-Reyes <kiko@erlang.org>
 - Charlie Jenkins <charlie@rivosinc.com>
 - Hatzka <hatzka@nezumi.studio>
+- Jonas Fierlings <fnoegip@gmail.com>
--- a/changelog.d/fixed/jujutsu-file-listing.md
+++ b/changelog.d/fixed/jujutsu-file-listing.md
@@ -1,0 +1,2 @@
+- Fixed repository file detection using jujutsu versions 0.19.0 or later.
+  (#1191)
--- a/src/reuse/vcs.py
+++ b/src/reuse/vcs.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: 2020 John Mulligan <jmulligan@redhat.com>
 # SPDX-FileCopyrightText: 2023 Markus Haug <korrat@proton.me>
 # SPDX-FileCopyrightText: 2024 Skyler Grey <sky@a.starrysky.fyi>
+# SPDX-FileCopyrightText: 2025 Jonas Fierlings <fnoegip@gmail.com>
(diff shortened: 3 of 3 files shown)


floriansnow:
I think your changes are good and I agree that we should usually support one version. However, in this case, version 0.19 was released only last July, so if a distro relies on the packed version, the deprecation may be too short notice. Perhaps it would make sense to put the version detection in for now and create an issue to remove it again at some point in the future?

PigeonF:
Looking at [repology](https://repology.org/project/jujutsu/versions), there are some distros shipping the outdated versions (i.e. jujutsu < 0.19.0) - most notably alpine 3.20. If you agree, I will put in the version detection to deal with both versions, and add a comment about removing the version detection in the future.

TheCodedProf:
CC: @Minion3665

Minion3665 (approved):
Thanks very much for fixing this - this looks good to me!

PigeonF:
In light of the recent release I have rebased the changes onto main (and added the missing entry to the `changelog.d` directory).

PR #1219, opened 2025-09-21
Bump Python to 3.10

carmenbianca:
Also remove the upper-bound for Python 4 in pyproject. It is completely unnecessary.

      changed files.


carmenbianca:
Running `pyupgrade --py310-plus` on all files would be cool, but it would conflict greatly with #1145.

PR #1239, opened 2025-10-03
Add support for py.typed files

euronion:
      changed files.


carmenbianca:
I believe `py.typed` should be an empty file, which `reuse lint` ignores anyway, but I'm happy to take this PR. Thank you!

PR #1240, opened 2025-10-03
License parse fix

### LINKED ISSUE #4: The tool currently very poorly deals with erroneous SPDX expressions
Given a file `erroneous-spdx.txt`:

```
SPDX-Copyright: Carmen

SPDX-License-Identifier: MIT OR BSD AND
```

The output of `reuse lint` is:

```
reuse._util - ERROR - Could not parse 'MIT OR BSD AND'
reuse.project - ERROR - erroneous-spdx.txt holds an SPDX expression that cannot be parsed, skipping the file
NO LICENSE

The following files have no license(s):
  erroneous-spdx.txt

NO COPYRIGHT

The following files have no copyright:
  erroneous-spdx.txt

SUMMARY

Bad licenses: 0
Missing licenses: 0
Unused licenses: 0
Used licenses: Apache-2.0, CC-BY-SA-4.0, CC0-1.0, GPL-3.0-or-later
Read errors: 0
Files with copyright information: 47 / 48
Files with license information: 47 / 48
```

The ERROR statements are just logger output from within the program. The file is then completely skipped over, and its (completely valid) SPDX-Copyright tag is ignored.

Is this sufficient, or should the plumbing somehow change to account for this edge case?

### LINKED ISSUE #1228: Re-do reading and parsing of files
Currently, `reuse` does approximately the following to lint a file:

1. Detect if a file is binary with the `binaryornot` library. If it is, don't parse it.
2. Read the first 4 kibibytes into memory.
3. If a REUSE snippet is in those 4 kibibytes, then read the entire file into memory.
4. Decode the file buffer as UTF-8 regardless of if it's actually UTF-8.
5. Run a regex on the file buffer to find relevant REUSE data.
6. Specifically for `SPDX-License-Identifier:` tags, if the value is invalid, then completely bork parsing the file. This is issue #4.

This is problematic for several reasons:

- The `is_binary` function call accounts for approximately 40% of runtime, which is not great.
- There could be REUSE information in the rest of the file which is never read.
- The 4 kibibytes could cut off in the middle of a line containing REUSE information, see #1213.
- The file may not be UTF-8, see #1218.
- Apparently the newline detection is broken, see #1226.

So I want to completely redo how `reuse` reads files into memory and does heuristics on it. My current proposal is simple:

1. Read 4 kibibytes (or some other random amount) of the file into memory.
2. Run heuristics on those 4 kibibytes:
  - Is it binary? If it is, don't continue. (This should hopefully speed up the `is_binary` function call by only running on a subset of the file instead of the entire file.)
  - What is its encoding?
  - What are its newlines?
3. Given the heuristical information, read the entire file into memory in chunks of 64 kibibytes (or some other random amount). Like #1213, do some cut-off at newlines to avoid chopped information. Doing this in chunks is important to prevent accidentally reading a 2 GiB file and loading it into memory.
4. Correctly decode each chunk given the heuristics from step 2.
5. Run regexes on each chunk to find REUSE data.
6. Deal better with `SPDX-License-Identifier:`, somehow. But this may be out-of-scope for this issue.

I want to do this in two stages. First, implement steps 3­–5 of my proposal, so that the entire file is read in chunks. Then, implement steps 1–2 in a separate PR.

_Hopefully_ the performance impact of this won't be huge, but I think it's acceptable to slow down if we can guarantee correct behaviour.

carmenbianca:
Fixes #4
Fixes #1228

- [x] Wrote tests.
      changed files.

### DIFF
--- a/changelog.d/added/invalid-expressions.md
+++ b/changelog.d/added/invalid-expressions.md
@@ -1,3 +1,3 @@
-- A new criterion 'Invalid expressions' has been added to `reuse lint`. Invalid
-  expressions are SPDX License Expressions which are not valid according to the
-  grammar of the SPDX specification. (#1240)
+- A new criterion 'Invalid SPDX License Expressions' has been added to
+  `reuse lint`. Invalid expressions are SPDX License Expressions which are not
+  valid according to the grammar of the SPDX specification. (#1240)
--- a/changelog.d/added/invalid-expressions.md
+++ b/changelog.d/added/invalid-expressions.md
@@ -1,0 +1,3 @@
+- A new criterion 'Invalid expressions' has been added to `reuse lint`. Invalid
+  expressions are SPDX License Expressions which are not valid according to the
+  grammar of the SPDX specification. (#1240)
--- a/changelog.d/changed/bad-licenses-per-file.md
+++ b/changelog.d/changed/bad-licenses-per-file.md
@@ -1,0 +1,3 @@
+- The 'Bad licenses' criterion in `reuse lint` previously searched for bad
+  licenses in every single file. Now, only bad licenses in `LICENSES/` are
+  detected, which is more in line with the documentation. (#1240)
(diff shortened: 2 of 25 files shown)


carmenbianca:
Invalid expressions still missing from summary at the end of `reuse lint`.

PR #1264, opened 2025-10-23
Support file-magic

### LINKED ISSUE #1261: reuse fails to work if python-file-magic is installed
The `python-magic` PyPI package conflicts with the upstream libmagic python bindings (which are not on PyPI). Unfortunately, even if another encoding detection dependency is selected by editing the `pyproject.toml` file, reuse will error out if the upstream bindings happen to be installed on the system:

```python
  File "/usr/bin/reuse", line 5, in <module>
    from reuse.cli.main import main
  File "/usr/lib64/python3.13/site-packages/reuse/cli/__init__.py", line 7, in <module>
    from . import (
    ...<8 lines>...
    )
  File "/usr/lib64/python3.13/site-packages/reuse/cli/annotate.py", line 30, in <module>
    from .._annotate import add_header_to_file
  File "/usr/lib64/python3.13/site-packages/reuse/_annotate.py", line 34, in <module>
    from .extract import contains_reuse_info
  File "/usr/lib64/python3.13/site-packages/reuse/extract.py", line 111, in <module>
    _MAGIC = get_encoding_module().Magic(mime_encoding=True)
TypeError: Magic.__init__() got an unexpected keyword argument 'mime_encoding'
```

This happens because both of the modules (the one from PyPI and the one from libmagic) are called `magic`, but do not have compatible API. The selection algorithm cannot differentiate between those two, and will try to use the incompatible version if that one happens to be present.

---

The `python-magic` PyPI package provides a compatibility layer ([docs](https://github.com/ahupp/python-magic/blob/master/COMPAT.md), [code](https://github.com/ahupp/python-magic/blob/07bd5dd0ed651465d086e57abc629b1071162ea3/magic/__init__.py#L479-L519)), so the most straightforward fix could be to "just" switch to using only the compat layer. However, the relevant functions are provided behind `PendingDeprecationWarning`, so presumably this might get removed at a future date.

Other option could be to extend the selection algorithm to poke around the API and try to determine which of the magic modules it actually has available.

I can probably submit a relevant PR, but I want to gather some feedback beforehand. Thoughts?

carmenbianca:
Fixes #1261 

- [x] Wrote tests.
      changed files.

### DIFF
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -733,6 +733,50 @@ class TestEncodingModule:
             " Aborting."
         )
 
+    def test_correct_env(self, encoding_module):
+        """If REUSE_ENCODING_MODULE is set to a valid value, use the correct
+        module.
+        """
+        env = os.environ.copy()
+        env["REUSE_ENCODING_MODULE"] = encoding_module
+        result = subprocess.run(
+            [
+                sys.executable,
+                "-c",
+                "import reuse.extract;"
+                "print(reuse.extract._get_encoding_module_name())",
+            ],
+            env=env,
+            stderr=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            check=False,
+        )
+        assert result.returncode == 0
+        assert result.stdout.decode("utf-8").strip() == encoding_module
+
+    def test_magic_env(self, encoding_module):
+        """If REUSE_ENCODING_MODULE is 'magic', import python-magic for
+        backwards compatibility.
+        """
+        if encoding_module != "python-magic":
+            pytest.skip("this test only works with python-magic")
+        env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -190,13 +190,23 @@ def multiprocessing(request, monkeypatch) -> Generator[bool, None, None]:
...
 
-@pytest.fixture(params=["magic", "charset_normalizer", "chardet"])
+@pytest.fixture(
+    params=["python-magic", "file-magic", "charset_normalizer", "chardet"]
+)
 def encoding_module(request, monkeypatch) -> Generator[bool, None, None]:
     """Run the test with or without libmagic."""
     try:
         if request.param == "magic" and not is_posix:
             pytest.skip("Windows not supported")
```
carmenbianca: if "magic" in request.param

PR #1277, opened 2025-11-21
Feature/1190 Use toml at annotate

### LINKED ISSUE #1190: don't ignore REUSE.toml in reuse annotate
We want to add missing annotations via a pre-commit hook (see #1178) but pre-commit is limited to pass either no filename or all filenames.
While some files have a annotation via `REUSE.toml`, `reuse annotate` will nevertheless add the copyright header.

→ It would be good to have an CLI argument to not ignore REUSE.toml or make this the default behavior.

Ka55i0peia:
Uses `REUSE.toml` at annotate run

Fixes #1190 

- [x] Wrote tests.
      changed files.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -160,3 +160,4 @@ Contributors
 - Thomas Gilon <thomas.gilon@openenergytransition.org>
 - Yongmin Hong <yewon@revi.email>
 - Martijn Saelens <https://github.com/MartenBE>
+- Jan Gietzel <jan.gietzel@gmail.com>


Ka55i0peia:
Thanks @carmenbianca  for your time diving into that!

What does the option `--import-global` do? (I haven't found references on that neither in doc nor in source.)

I guess what’s intended by #1190 is actually what you describe as feature 2.

Regarding the use case you mentioned in feature 1: in my eyes, when reuse annotate tries to annotate a file that is already listed in `REUSE.toml`, it should update the annotation inside the file (or, depending on `--skip-existing`, simply ignore the file). That behaviour is imho covered by the current code.

Your point about the case where a file is not present in `REUSE.toml` when calling `reuse annotate file1 file2` is interesting and important. In my opinion, the command **should report *all* files for which such an annotation is missing**. That way, the user can either re-run `reuse annotate file1 file2` and adding parameters `--copyright` and `--license`, or add the necessary annotations to `REUSE.toml`.
:bulb: In the current implementation of this PR here, the command fails with _the first file_ and terminates. (So the user have to iterate like this through all missing annotations, which is maybe annoying and can be avoided by reporting all files where an annotation is missing.)

Same might apply when command `reuse annotate` (without files or copyright options on cli) hits a file, that is not listed in TOML. Suggestion: Warn/report all files where an annotation is missing. (Maybe an option like `--skip-missing` would be useful to silence pass that.)

> What does reuse `annotate --year 2025 file1` mean, for instance?

In my eyes it shall annotate `file1` using config from toml and overwriting annotion with year 2025 (when specified in toml differently).

This is how I understand the command `annotate` of `reuse`. Apologies if I am getting design or plans wrong for that tool. 

Please let me know what you think on this.

I like this idea of reuse annote so it can be used easily in pre-commit hook. Thanks for this great tool! 👍

carmenbianca:
> What does the option --import-global do? (I haven't found references on that neither in doc nor in source.)

It's my proposal for a flag to enable feature 2. It does not presently exist.

> I guess what’s intended by https://github.com/fsfe/reuse-tool/issues/1190 is actually what you describe as feature 2.

The other way around. #1190 describes feature 1. This PR implements feature 2, unless I am mistaken.

> Regarding the use case you mentioned in feature 1: in my eyes, when reuse annotate tries to annotate a file that is already listed in REUSE.toml, it should update the annotation inside the file (or, depending on --skip-existing, simply ignore the file). That behaviour is imho covered by the current code.

The tool should not apply information from the REUSE.toml file to the file without explicit instructions to do so. Or: `reuse annotate` should **only** apply the information included in its CLI arguments unless instructed otherwise.

`--skip-existing` does not currently implement feature 1, I believe. I'm not certain. I believe it checks whether there is REUSE information _inside of the file_, and skips if there is.

> In my opinion, the command should report all files for which such an annotation is missing.

Sure. I can see that being useful.

> Same might apply when command reuse annotate (without files or copyright options on cli) hits a file, that is not listed in TOML

This shouldn't happen. Associating information with files in REUSE.toml is the exception, not the rule. REUSE.toml is not a repetition of all licensing in the project—it exists to facilitate cases where putting information inside of files is impossible, inconvenient, or undesirable. Listing annotated files that don't appear in REUSE.toml would—in an ideal repository—flag every single file. And that wouldn't be very helpful. At best, if this information is helpful, we can log it so you can see it with `reuse --debug [...]`.

> In my eyes it shall annotate file1 using config from toml and overwriting annotion with year 2025 (when specified in toml differently).

I am not certain I agree with this. I will reiterate an earlier point:

Neither feature 1 nor feature 2 should be default behaviour. They should only happen when calling with `reuse annotate --feature1` (probably `reuse annotate --skip-existing` or `reuse annotate --skip-global`) or `reuse annotate --feature2` (dubbed `reuse annote --import-global`, but we can name it something else).

Feature 2 is a significant departure from the default behaviour of `reuse annotate`—it copies information from REUSE.toml into the specified files, rather than putting information defined as arguments to `reuse annotate` into the specified files. It almost definitely shouldn't modify the information in transit.

I am also not certain if feature 2 should be used _simultaneously_ with the regular `annotate` workflow. `reuse annotate --feature2 --copyright Jane Doe file1` possibly seems like a bad idea to me. Logically, it would do two things simultaneously:

- Apply information from REUSE.toml into file1;
- and insert `SPDX-FileCopyrightText: 2025 Jane Doe` into file1.

`reuse annotate --feature2 --copyright Jane Doe --year 2017 file1`, then, has two ways of being interpreted:

1. Apply information from REUSE.toml into file1 as-is and insert `SPDX-FileCopyrightText: 2017 Jane Doe` into file1;
2. or apply information from REUSE.toml into file1 with all years changed to '2017' and also insert `SPDX-FileCopyrightText: 2017 Jane Doe` into file1.

Option 2 seems like a really bad idea to me, because it means that `--copyright` and `--year` apply differently.

But option 1 is still a little opaque. I don't know if there is a real use-case for doing these two operations simultaneously. It's probably fine, but I'm not sure.

One option is to put feature 2 into a completely different subcommand, e.g. `reuse copy-from-global` (or whatever), but there is heaps of logic in `reuse annotate` that we would want to re-use, so it doesn't really make sense to make it a different subcommand.

---

As an important aside: Feature 1 and feature 2 are not interdependent. One can be implemented without the other. In fact, the two features are mutually exclusive.

